PR #9, opened 2021-11-19
test: add unit tests

dongheeleeme:
Hi :) I added missing unit tests for `isArray` and `isString` since It's not coved by unit tests. Please review on it. Thanks!

### DIFF
--- a/test/isArray.spec.ts
+++ b/test/isArray.spec.ts
@@ -1,7 +1,7 @@
 import { isArray } from "../src/index";
 
-describe("isString", function () {
-  it.each([2, true, "a"])(
+describe("isArray", function () {
+  it.each([undefined, null, true, 1, "a", Symbol("a"), () => null])(
     "given non array then should return false",
     function (s) {
       const result = isArray(s);
--- a/test/isArray.spec.ts
+++ b/test/isArray.spec.ts
@@ -1,14 +1,15 @@
 import { isArray } from "../src/index";
 
-describe("isString", function() {
-  it.each(
-    [2, true, "a"]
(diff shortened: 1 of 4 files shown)


ppeeou:
@dongheeleeme Thank you for your work:) could you check `prettier`?? you can try `npm run prettier:fix`

ppeeou:
@dongheeleeme Could you add a type test as well?

you can see the reference
https://github.com/marpple/FxTS/tree/main/type-check

dongheeleeme:
Thanks @ppeeou! Updated in 0135574
Btw, do we have a way to run `type-check` files easily? Honestly, I don't have much idea about `type-check` directory.

test/isArray.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { isArray } from "../src/index";
+
+describe("isString", function () {
+  it.each([2, true, "a"])(
```
ppeeou, line 4: It would be better to put all types of javascript
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#description

`isString` too

ppeeou:
you can try `npm run compile:check`. we would write this README 😀
https://github.com/marpple/FxTS/commit/40fb0fc4db07170174618a5af87bde073be68715

test/isArray.spec.ts
dongheeleeme: Thanks, I added all of the js types in here(2f2d020) 👍

ppeeou (approved):
LGTM

shine1594 (approved):

PR #10, opened 2021-11-20
feat: add overloading type of 'filter' function

jbl428:
This PR is to make better type inference of filter function when boolean constructor is used.

```ts
const res = filter(Boolean, [undefined, null, 0, "", false, -0, 123] as const) // IterableIterator<123> 
```


shine1594:
LGTM

ppeeou (approved):
LGTM 👍

PR #11, opened 2021-11-20
feat: add type predicate overloading of 'filter' function

jbl428:
To make `type predicate` work as expected in filter function.
I forgot to add this one in previous PR 😄 

```ts

declare function isString(s: unknown): s is string;

const res5 = filter(isString, [1, "a", true]); // IterableIterator<string>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
       "prettier --write",
       "git add"
     ],
-    "*.json": [
+    "*.(json|md)": [
       "prettier --write",
       "git add"
     ]
--- a/src/Lazy/uniqBy.ts
+++ b/src/Lazy/uniqBy.ts
@@ -92,14 +92,14 @@ function uniqBy<A extends Iterable<unknown> | AsyncIterable<unknown>, B>(
 
   if (isIterable(iterable)) {
     return pipe(
-      iterable as Iterable<IterableInfer<A>>,
+      iterable,
       filter((a) => pipe1(f(a), checkAndAdd)),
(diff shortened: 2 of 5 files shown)


src/Lazy/uniqBy.ts
```diff
@@ -92,14 +92,14 @@ function uniqBy<A extends Iterable<unknown> | AsyncIterable<unknown>, B>(
 
   if (isIterable(iterable)) {
     return pipe(
-      iterable as Iterable<IterableInfer<A>>,
+      iterable,
       filter((a) => pipe1(f(a), checkAndAdd)),
```
jbl428, line 96: I change this line to remove type error.
It was caused by following definition.

```ts
function filter<
  A extends Iterable<unknown> | AsyncIterable<unknown>,
  B extends IterableInfer<A>,
>(
  f: (a: IterableInfer<A>) => a is B,
): (iterable: A) => ReturnIterableIteratorType<A, B>;
```

I can not sure it is right definition.

package.json
```diff
@@ -85,7 +85,7 @@
       "prettier --write",
       "git add"
     ],
-    "*.json": [
+    "*.(json|md)": [
```
jbl428, line 88: pre commit hook did not check markdown files but github action does.
so CI failed if someone edit markdown file and did not run prettier.

ppeeou (approved):
We didn't think of this part Thank you for your contribution👍

shine1594 (approved):
Thank you

PR #14, opened 2021-11-22
fix: missed signature in reduce function

### LINKED ISSUE #13: Missed signature in `reduce` function
# Bug Report

### 💻 Code

```ts
const work = reduce((a, b) => a + b, [1, 2, 3]);
const notWork = reduce((a, b) => a + b, toAsync([1, 2, 3]));
```

### 🙁 Actual behavior

There is a problem that the value of `a` is inferred as `unknown` when `AsyncIterable` type is passed as an argument.
This bug is probably caused by missing some signatures of the `reduce` function.

### 🙂 Expected behavior

It would be nice to be able to infer the value `a` in `AsyncIterable` normally.

---

### Version Infomation

- Operating system: Mac 12.0.1
- Typescript: 4.5.0
- Nodejs: 14.18.1

6h15m:
Fixes #13

### DIFF
--- a/src/reduce.ts
+++ b/src/reduce.ts
@@ -82,6 +82,11 @@ function reduce<A, B>(f: (a: B, b: A) => B, iterable: Iterable<A>): B;
 
 function reduce<A, B>(f: (a: B, b: A) => B, seed: B, iterable: Iterable<A>): B;
 
+function reduce<A>(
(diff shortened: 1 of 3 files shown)


shine1594 (approved):
LGTM 👍

PR #32, opened 2021-12-01
feat: add zipWith function

### LINKED ISSUE #31: Add `zipWith` function
# Suggestion

## ⭐ Suggestion

Returns Iterable/AsyncIterable out of the two supplied by applying `f` to each same positioned pair in Iterable/AsyncIterable.

## 💻 Use Cases

```ts

const iter = zipWith((a,b) => [a,b], [1,2,3], ['a','b','c']);
iter.next(); // {value: [1, 'a'] , done: false}
iter.next(); // {value: [2, 'b'] , done: false}
iter.next(); // {value: [3, 'c'] , done: false}

```

ppeeou:
Fixes #31

we do not support `curry` in `zipWith` not yet

### DIFF
--- a/src/Lazy/zipWith.ts
+++ b/src/Lazy/zipWith.ts
@@ -54,7 +54,7 @@ function zipWith<A, B, C>(
   }
 
   throw new TypeError(
-    "'iterable1' 'iterable2' must be type of Iterable or AsyncIterable",
+    "'iterable1' and 'iterable2' must be type of Iterable or AsyncIterable",
   );
 }
 
--- a/src/Lazy/zipWith.ts
+++ b/src/Lazy/zipWith.ts
@@ -1,0 +1,61 @@
+import { map, zip } from ".";
+import { isAsyncIterable, isIterable } from "../_internal/utils";
+
+/**
+ * Returns Iterable/AsyncIterable out of the two supplied by applying `f` to each same positioned pair in Iterable/AsyncIterable.
+ *
+ * @example
+ * const iter = zipWith((a,b) => [a,b], [1,2,3], ['a','b','c']);
+ * iter.next(); // {value: [1, 'a'] , done: false}
+ * iter.next(); // {value: [2, 'b'] , done: false}
+ * iter.next(); // {value: [3, 'c'] , done: false}
+ * iter.next(); // {value: undefined , done: true}
+ */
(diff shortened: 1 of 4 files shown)


src/Lazy/zipWith.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+  if (isAsyncIterable(iterable1) && isIterable(iterable2)) {
+    return map(([a, b]) => f(a, b), zip(iterable1, iterable2));
+  }
+  if (isAsyncIterable(iterable1) && isAsyncIterable(iterable2)) {
+    return map(([a, b]) => f(a, b), zip(iterable1, iterable2));
+  }
+
+  throw new TypeError(
+    "'iterable1' 'iterable2' must be type of Iterable or AsyncIterable",
```
shine1594, line 57:
```suggestion
    "'iterable1' and 'iterable2' must be type of Iterable or AsyncIterable",
```

src/Lazy/zipWith.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+  if (isIterable(iterable1) && isAsyncIterable(iterable2)) {
+    return map(([a, b]) => f(a, b), zip(iterable1, iterable2));
+  }
+  if (isAsyncIterable(iterable1) && isIterable(iterable2)) {
+    return map(([a, b]) => f(a, b), zip(iterable1, iterable2));
+  }
+  if (isAsyncIterable(iterable1) && isAsyncIterable(iterable2)) {
+    return map(([a, b]) => f(a, b), zip(iterable1, iterable2));
+  }
```
shine1594, line 54: The current code makes the intent a little clearer, but I think it's fine to keep it concise in this way.
```suggestion
  if (isAsyncIterable(iterable1)) {
    return map(([a, b]) => f(a, b), zip(iterable1, iterable2));
  }
  if (isAsyncIterable(iterable2)) {
    return map(([a, b]) => f(a, b), zip(iterable1, iterable2));
  }
```

ppeeou: You are right. However, I think it is right to check TypeError in all situations because it is a code written to notify in runtime situations

shine1594: Okay, this is good enough.

shine1594 (approved):

PR #33, opened 2021-12-02
chore: set custom variable

ppeeou:
Users may be confused about the error phrase, so it is displayed in a custom variable -> ''


ppeeou:
@hg-pyun thank you for your report

PR #34, opened 2021-12-02
fix: consume iterable as many as call in takeUntil

ppeeou:
we did not test like this

```ts
it("should be able to take the element", async function () {
  const fn = jest.fn();
  const res = await pipe(
    toAsync([3, 3, 1, 0, 0]),
    peek(fn),
    map((a) => delay(100, a)),
    takeUntil((a) => a < 3),
    toArray,
  );

  expect(fn).toHaveBeenCalledTimes(3);
  expect(res).toEqual([3, 3, 1]);
}, 350);
```

### DIFF
--- a/src/Lazy/takeUntil.ts
+++ b/src/Lazy/takeUntil.ts
@@ -30,6 +30,10 @@ function asyncSequential<A, B>(
       return this;
     },
     async next(_concurrent) {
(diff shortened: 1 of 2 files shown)


ppeeou:
move https://github.com/marpple/FxTS/pull/35

PR #48, opened 2021-12-07
feat: add includes function

### LINKED ISSUE #46: Add `includes` function
# Suggestion

## ⭐ Suggestion

Checks if the specified value is equal.

## 💻 Use Cases

```ts
includes('c', 'abcd'); // true
includes(3, [1,2,3,4]); // true
```

ppeeou:
Fixes #46

### DIFF
--- a/src/includes.ts
+++ b/src/includes.ts
@@ -1,0 +1,54 @@
+import { some } from ".";
+import ReturnValueType from "./types/ReturnValueType";
+import { isAsyncIterable, isIterable } from "./_internal/utils";
+
+/**
+ * Checks if the specified value is equal. (contains)
+ *
+ * @example
+ * ```ts
+ * includes('c', 'abcd'); // true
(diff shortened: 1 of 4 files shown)


src/includes.ts
```diff
@@ -0,0 +1,63 @@
...
+    return (iterable) => {
+      return includes(value, iterable as any) as ReturnValueType<T, boolean>;
+    };
+  }
+
+  if (isIterable(iterable)) {
+    return pipe(
+      iterable as Iterable<IterableInfer<T>>,
+      find((a) => a === value),
+      (a) => a !== undefined,
```
shine1594, line 48: How about implementing using `some` like this?

```suggestion
    return some(a => a === value, iterable);
```

ppeeou: oh yes your right :)

shine1594 (approved):

PR #55, opened 2021-12-08
feat: add 'isObject' function

### LINKED ISSUE #51: Add `isObject` function
# Suggestion

## ⭐ Suggestion

https://lodash.com/docs/4.17.15#isObject

Since the purpose of this function is to check only objects, not primitive values, it seems to be implemented in the same way as isObject of lodash.

## 💻 Use Cases

https://lodash.com/docs/4.17.15#isObject

shine1594:
Fixes #51


ppeeou (approved):
LGTM

PR #58, opened 2021-12-08
fix: handle tap async

ppeeou:
why?

If the auxiliary function of tap is asynchronous, it is more natural to wait and return the original value.


shine1594 (approved):
👍

PR #89, opened 2021-12-14
feat: add `consume` function

### LINKED ISSUE #66: Add `consume` function
# Suggestion

## ⭐ Suggestion

Consumes the given number of `Iterable/AsyncIterable`. If the number is empty, all is consumed.

## 💻 Use Cases

```ts
const iterator = (function *(){ 
  yield 1; 
  yield 2; 
  yield 3;  
})();
consume(iterator, 2);
iterator.next(); // {value:3, done:false}

// with pipe
pipe(
  range(10),
  peek(updateApi),
  concurrent(5),
  consume,
);

```

ppeeou:
Fixes #66

### DIFF
--- a/src/consume.ts
+++ b/src/consume.ts
@@ -1,0 +1,68 @@
+import ReturnValueType from "./types/ReturnValueType";
+import { isAsyncIterable, isIterable } from "./_internal/utils";
+import range from "./Lazy/range";
+
+function sync<T>(iterable: Iterable<T>, n: number) {
+  const iterator = iterable[Symbol.iterator]();
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for (const _ of range(0, n)) {
+    if (iterator.next().done) {
+      return;
+    }
+  }
+}
+
+async function async<T>(iterable: AsyncIterable<T>, n: number) {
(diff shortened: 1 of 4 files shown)


src/consume.ts
```diff
@@ -0,0 +1,70 @@
+import ReturnValueType from "./types/ReturnValueType";
+import { isAsyncIterable, isIterable } from "./_internal/utils";
+import range from "./Lazy/range";
+
+function sync<T>(iterable: Iterable<T>, n: number) {
+  const iterator = iterable[Symbol.iterator]();
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for (const _ of range(0, n)) {
+    const { done } = iterator.next();
+    if (done) {
```
shine1594, line 10:
```suggestion
    if (iterator.next().done) {
```

test/consume.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { consume, peek, pipe, toAsync } from "../src/index";
+
+describe("each", function () {
```
shine1594, line 3: This should be changed to 'consume'.

shine1594 (approved):

PR #92, opened 2021-12-15
fix: change wrong pattern for test

khg0712:
I found tests of FxTs are consist of for type check and for unit test.
But all type check only tests run and are failed when I run `npm test`.

Your test script specified glob pattern of under `/test` directory. But jest does't recognize it.

So, I investigated about it. and I found an issue about this. ([comment](https://github.com/facebook/jest/issues/7108#issuecomment-604905294)). and I referred to the comment and changed your config.
I'm still confused why jest doesn't work with your config. But I guess It's because of micromatch which is used for glob pattern.

BEFORE:
all type check tests are failed
image

AFTER:
all type check tests are not target.
image

ps. Sometimes some tests are failed because they depend on timeout api. I think it'd be nice to fix it. thx 
image

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -1,4 +1,4 @@
-const path = require('path');
+const path = require("path");
 
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -1,5 +1,8 @@
+const path = require('path');
+
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
   preset: "ts-jest",
   testEnvironment: "node",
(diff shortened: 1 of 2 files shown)


ppeeou:
@khg0712  Thank you for your contribution
Could you check `prettier`?

ppeeou:
> ps. Sometimes some tests are failed because they depend on timeout api. I think it'd be nice to fix it. thx

also registered it https://github.com/marpple/FxTS/issues/93 👍

khg0712:
@ppeeou I'll fix it :)

khg0712:
> @khg0712  Thank you for your contribution
> Could you check `prettier`?
> 
> you can try `npm run prettier:fix`

@ppeeou I'll fix it :)

shine1594 (approved):
Thank you 🙂

PR #95, opened 2021-12-16
feat: add `cycle` function

### LINKED ISSUE #78: Add `cycle` function
# Suggestion

## ⭐ Suggestion

ref https://docs.python.org/3/library/itertools.html#itertools.cycle

## 💻 Use Cases

```ts
const iter = cycle("abc")
iter.next(); // {value:"a", done: false}
iter.next(); // {value:"b", done: false}
iter.next(); // {value:"c", done: false}
iter.next(); // {value:"a", done: false}
iter.next(); // {value:"b", done: false}

// with pipe
pipe(
  cycle([1,2,3,4]),
  take(5),
  toArray,
); // [1,2,3,4,1]


```

ppeeou:
Fixes #78


shine1594 (approved):
LGTM

PR #99, opened 2021-12-20
feat: add `repeat` function

### LINKED ISSUE #60: Add `repeat` function
# Suggestion

## ⭐ Suggestion

Returns a Iterable/AsyncIterable of size `n` containing  a specified value.

## 💻 Use Cases

```ts
const iter = repeat(2, 10);
iter.next(); // {value: 10, done:false}
iter.next(); // {value: 10, done:false}
iter.next(); // {value: undefined, done:true}

// with pipe
pipe(
  repeat(2, 10),
  toArray,
); // [10, 10]

```

ppeeou:
Fixes #60

### DIFF
--- a/src/Lazy/index.ts
+++ b/src/Lazy/index.ts
@@ -16,6 +16,7 @@ import pluck from "./pluck";
 import prepend from "./prepend";
 import range from "./range";
 import reject from "./reject";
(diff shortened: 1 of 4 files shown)


shine1594 (approved):
LGTM

PR #102, opened 2021-12-20
feat: add max/min function

### LINKED ISSUE #79: Add `min` and `max` functions
# Suggestion

## 💻 Use Cases

```ts
min([5, 3, 1]); // 1;
min([5, NaN, 1]); // NaN;

max([1, 3, 5]); // 5;
max([1, NaN, 1]); // NaN;
max([1, Infinity, 1]); // Infinity;

await min(toAsync([5, 3, 1])); // 1;
await max(toAsync([1, 3, 5])); // 5;
```

ppeeou:
Fixes #79

### DIFF
--- a/src/max.ts
+++ b/src/max.ts
@@ -1,0 +1,66 @@
+import ReturnValueType from "./types/ReturnValueType";
+import { isAsyncIterable, isIterable } from "./_internal/utils";
+
+function sync(iterable: Iterable<number>) {
+  let n = NaN;
+  for (const a of iterable) {
+    if (Number.isNaN(a)) {
+      return a;
+    } else if (Number.isNaN(n)) {
+      n = a;
+    } else if (a > n) {
+      n = a;
+    }
+  }
+
+  if (Number.isNaN(n)) {
+    return 0;
+  }
+
+  return n;
+}
+
+async function async(iterable: AsyncIterable<number>) {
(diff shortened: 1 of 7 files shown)


src/max.ts
```diff
@@ -0,0 +1,66 @@
...
+    } else if (Number.isNaN(n)) {
+      n = a;
+    } else if (a > n) {
+      n = a;
+    }
+  }
+
+  if (Number.isNaN(n)) {
+    return 0;
+  }
```
shine1594, line 37: `Math.max()` returns `-Infinity`. Therefore, it seems appropriate that this code also returns `-Infinity`.

src/max.ts
```diff
@@ -0,0 +1,66 @@
...
+function sync(iterable: Iterable<number>) {
+  let n = NaN;
+  for (const a of iterable) {
+    if (Number.isNaN(a)) {
+      return a;
+    } else if (Number.isNaN(n)) {
+      n = a;
+    } else if (a > n) {
+      n = a;
+    }
```
shine1594, line 13: Since in most cases there will be no `NaN' value,
It looks a little better to write it like this:
```suggestion
    }
    if (a > n || Number.isNaN(n)) {
      n = a;
    }
```

shine1594 (approved):
👍

PR #107, opened 2021-12-22
feat: add `size` function

### LINKED ISSUE #75: Add `size` function
# Suggestion

## 💻 Use Cases

```ts
size(range(10, 16)); // 5
await size(toAsync(range(10, 16))); // 5
```

ppeeou:
Fixes #75

### DIFF
--- a/src/size.ts
+++ b/src/size.ts
@@ -1,0 +1,45 @@
+import each from "./each";
+import ReturnValueType from "./types/ReturnValueType";
+import { isAsyncIterable, isIterable } from "./_internal/utils";
+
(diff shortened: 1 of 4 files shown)


src/size.ts
```diff
@@ -0,0 +1,45 @@
+import { each } from ".";
```
shine1594, line 1: In the current build process, there is a step to attach the `.js` extension, so it should be used as follows.
```suggestion
import each from "./each";
```

ppeeou: thank you for your comment :) I changed it!

shine1594 (approved):

PR #110, opened 2021-12-22
Add `juxt` function

### LINKED ISSUE #61: Add `juxt` function
# Suggestion

## ⭐ Suggestion

Applies a list of function to a `Iterable/AsyncIterable` of values.

## 💻 Use Cases

```ts
juxt([Math.max, Math.min])(1,2,3,-4,5,6) // [6, -4]
```

shine1594:
Fixes #61

### DIFF
--- a/src/juxt.ts
+++ b/src/juxt.ts
@@ -2,8 +2,7 @@ import type Append from "./types/Append";
 import type Arrow from "./types/Arrow";
 import type Cast from "./types/Cast";
 import type Tail from "./types/Tail";
-
-type HasItem<T extends any[]> = T extends [any, ...any] ? true : false;
+import type HasItem from "./types/HasItem";
 
 type ParameterTuples<FS extends Arrow[], T = []> = HasItem<FS> extends true
   ? ParameterTuples<Tail<FS>, Append<Cast<T, any[]>, Parameters<FS[0]>>>
--- a/src/juxt.ts
+++ b/src/juxt.ts
@@ -26,6 +26,29 @@ type JuxtReturnTypes<ARGS, FS extends Arrow[]> = ARGS extends undefined
   ? void
   : ReturnTuples<FS>;
 
+/**
+ * `juxt` applies a list of functions to a list of values.
+ *
+ * @example
+ * ```ts
+ * const range = juxt([Math.min, Math.max], [1, 2, 3, 4]); // [1, 4]
+ *
+ * // with pipe
+ * const entries = (obj: { a: number; b: number }) =>
+ *   pipe(
+ *     [Object.keys, Object.values] as const,
+ *     juxt,
+ *     (f) => f(obj),
+ *     apply(zip),
+ *     toArray,
+ *   );
+ *
+ * entries({ a: 1, b: 2 }); // [ ["a", 1], ["b", 2] ]
+ * ```
+ *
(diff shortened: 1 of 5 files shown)


src/juxt.ts
```diff
@@ -0,0 +1,58 @@
...
+ * entries({ a: 1, b: 2 }); // [ ["a", 1], ["b", 2] ]
+ * ```
+ *
+ * see {@link https://fxts.dev/docs/pipe | pipe}, {@link https://fxts.dev/docs/apply | apply}
+ */
+
+function juxt<FS extends Arrow[]>(
+  fs: [...FS] | readonly [...FS],
+): (...args: JuxtArgs<FS>) => JuxtReturnTypes<JuxtArgs<FS>, FS> {
+  return (...args: JuxtArgs<FS>) => fs.map((f: Arrow) => f(...args)) as any;
```
ppeeou, line 55: Doesn't juxt provide asynchronous functions?

shine1594: I think the user should be able to choose whether to evaluate asynchronous results (promises) sequentially or concurrently.

```ts

const asyncJob = (a: number, b: number) => delay(1000, a + b);

pipe (
  [1, 2],
  apply(juxt([asyncJob, asyncJob, asyncJob])),
  toAsync,
  concurrent(3),
  toArray,
);
```

type-check/juxt.test.ts
```diff
@@ -0,0 +1,34 @@
+import * as Test from "../src/types/Test";
+import { apply, pipe, juxt } from "../src";
+
+const { checks, check } = Test;
+
+declare function f1(a: number, b: string): string;
+declare function f2(a: number, b: string): boolean;
+declare function f3(a: number, b: boolean): number;
+declare function f4(a: number, b: string, c: boolean): number;
```
ppeeou, line 9:
```suggestion
// declare function f4(a: number, b: string, c: boolean): number;
```

how about comment `//` ?

type-check/juxt.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const res1 = juxt([])();
+const res2 = juxt([f1, f2])(1, "2");
+
+// pipeline
+const res3 = pipe([1, 2, 3, 4], apply(juxt([Math.max, Math.min])));
+const res4 = pipe(juxt([f1, f2]), (f) => f(1, "2"));
+const res5 = pipe([f1, f2] as const, juxt, (f) => f(1, "2"));
+const res6 = pipe([f1, f2] as [typeof f1, typeof f2], juxt, (f) => f(1, "2"));
```
ppeeou, line 18: 👍

src/juxt.ts
```diff
@@ -0,0 +1,58 @@
+import type Append from "./types/Append";
+import type Arrow from "./types/Arrow";
+import type Cast from "./types/Cast";
+import type Tail from "./types/Tail";
+
+type HasItem<T extends any[]> = T extends [any, ...any] ? true : false;
```
ppeeou, line 6: It may be good to create a separate file

ppeeou (approved):
LGTM

type-check/juxt.test.ts
shine1594: That would be better.

src/juxt.ts
shine1594: The `HasItem` type has been separated into a separate file.

PR #111, opened 2021-12-22
fix: handle sum type error

ppeeou:
fix it to occur an error below situation
const res0 = sum([1, 'a']); // error


shine1594 (approved):

hg-pyun (approved):
LGTM

PR #112, opened 2021-12-23
Change evaluation method to lazy evaluation on `reverse` function

ppeeou:
why?
in `array` case, lazy evaluation makes sense.


shine1594 (approved):
👍

PR #118, opened 2021-12-24
feat: add `average` function

### LINKED ISSUE #77: Add `average` function
# Suggestion

## 💻 Use Cases

```ts
average([1,2,3,4,5]); // 3
await average(toAsync([1,2,3,4,5])); // 3
```

ppeeou:
Fixes #77

### DIFF
--- a/src/average.ts
+++ b/src/average.ts
@@ -1,0 +1,48 @@
+import { peek, pipe, sum } from ".";
+import ReturnValueType from "./types/ReturnValueType";
+import { isAsyncIterable, isIterable } from "./_internal/utils";
+
+/**
+ * Returns the average of the given (Iterable/AsyncIterable)<number> (mean)
+ *
+ * @example
+ * ```ts
+ * average([1, 2, 3, 4, 5]); // 6
+ * await average(toAsync([1, 2, 3, 4, 5])); // 6
+ *
+ * // with pipe
+ * pipe(
+ *  [1, 2, 3, 4, 5],
+ *  average,
+ * ); // 6
+ * ```
+ *
+ * see {@link https://fxts.dev/docs/pipe | pipe}
+ */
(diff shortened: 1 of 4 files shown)


src/average.ts
```diff
@@ -0,0 +1,48 @@
...
+  iterable: T,
+): ReturnValueType<T> {
+  let size = 0;
+
+  if (isIterable(iterable)) {
+    return pipe(
+      iterable as Iterable<number>,
+      peek(() => size++),
+      sum,
+      (a) => (size === 0 ? NaN : a / size),
```
hg-pyun, line 32: I think this part is the same as the logic below. Is there any way to combine it into one?

hg-pyun (approved):
It's a minor comment, so I approved.

src/average.ts
```diff
@@ -0,0 +1,48 @@
+import { peek, pipe, sum } from ".";
```
shine1594, line 1:
```suggestion
import peek from "./peek";
import pipe from "./pipe";
import sum from "./sum";
```

src/average.ts
ppeeou: Do you want to create it as a function?
like this..? `const divide = (a: number) => size === 0 ? NaN : a / size;`

shine1594: I guess that @hg-pyun says something like this:
```ts
if (isIterable(iterable) || isAsyncIterable(iterable)) {
  return pipe(
    iterable as Iterable<number> | AsyncIterable<number>,
    peek(() => size++),
    sum,
    (a) => (size === 0 ? NaN : a / size),
  ) as ReturnValueType<T>;
}
```

shine1594 (approved):

src/average.ts
ppeeou: We can also combine after this work
I think we can merge it first and then edit everything later.

PR #122, opened 2021-12-31
fix: change return type of `isIterable`

jbl428:
very minor change 

change return type of `isIterable` util function for better type inference.

### DIFF
--- a/src/Lazy/cycle.ts
+++ b/src/Lazy/cycle.ts
@@ -79,11 +79,11 @@ function cycle<T extends Iterable<unknown> | AsyncIterable<unknown>>(
 ):
   | IterableIterator<IterableInfer<T>>
   | AsyncIterableIterator<IterableInfer<T>> {
-  if (isIterable(iterable)) {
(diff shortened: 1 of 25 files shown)


ppeeou:
I left a comment

If this part is applied, how about changing other function files as well?

ex) some.ts
```ts
 if (isIterable<IterableInfer<A>>(iterable)) {
    return pipe(
      map(f, iterable),
      takeUntil(identity),
      reduce((a, b) => a || b),
      Boolean,
    );
  }

  if (isAsyncIterable<IterableInfer<A>>(iterable)) {
    return pipe(
      map(f, iterable),
      takeUntil(identity),
      reduce((a, b) => a || b),
      Boolean,
    );
  }
```

jbl428:
I think it is a great idea. :)
I will change other files.

ppeeou (approved):
good, I think it seems to be better :)

shine1594 (approved):
It seems to be a more readable code 👍

hg-pyun (approved):
LGTM

PR #124, opened 2022-01-07
fix: support async tap when curried

ppeeou:
It should support asynchronous function when `tap` is curried


src/tap.ts
```diff
@@ -31,9 +31,7 @@ function tap<T, U>(
   | Promise<T>
   | ((v: T) => U extends Promise<any> ? Promise<Awaited<T>> : T) {
   if (v === undefined) {
-    return (v: T) =>
-      (v instanceof Promise ? v.then(f) : f(v as Awaited<T>),
```
ppeeou, line 35: f(v as Awaited<T>) 
It have to wait for this result.

shine1594 (approved):

Issue #93, opened 2021-12-16
`npm run test` failed intermittently due to timeout


ppeeou:
# Bug Report

### 💻 Code

`npm run test` failed intermittently due to timeout.

![image](https://user-images.githubusercontent.com/25566139/146245134-29e0c889-9b97-41fa-8004-45ec9ecc83fe.png)

### 🙁 Actual behavior

The `setTimeout` set in test does not match intermittently.

### 🙂 Expected behavior

The test action should always succeed.

---

### Version Infomation

- Operating system:
- Typescript:
- Nodejs: 16.3.0

ppeeou:
We do a lot of testing asynchronous operations, and we want them to work as we intended by the correct time.

We can fail if we run jest in parallel.
execution environment
- branch: feature/jest-dockerenv
```
docker run --cpus 2 --rm -v `PWD`:/app --name fxts -it fxts:0.1 bash
npm test
```

No failure with jest maxWorkers set to 1
https://jestjs.io/docs/cli#--maxworkersnumstring

ppeeou:
In the past, the case of failure because it took a long time does not occur, but other cases are occurring
![image](https://user-images.githubusercontent.com/10924072/147042160-fafb36fc-3713-4dc7-b78d-d1c72d7c671f.png)

ppeeou:
https://jestjs.io/docs/timer-mocks

> Timer Mocks
The native timer functions (i.e., setTimeout, setInterval, clearTimeout, clearInterval) are less than ideal for a testing environment since they depend on real time to elapse. Jest can swap out timers with functions that allow you to control the passage of time. Great Scott!


We measure maximum and minimum times to measure time.

However, we are using `setTimeout` for the minimum time, but `setTimeout` does not seem to be suitable as a test.

https://github.com/marpple/FxTS/blob/main/test/utils.ts#L3-L5
```ts
export const callFuncAfterTime = (callback: Arrow, time = 1000) => {
  setTimeout(callback, time);
};
```

sunrabbit123:
All tests must be passed, but some tests fail depending on the environment.
Of course, maxWorkers is 1.

![image](https://user-images.githubusercontent.com/64676070/159845384-35a096a6-73cc-4a41-b8ae-95120e32f041.png)

```
      × should be chunked after concurrent (655 ms) // around 650ms                                                                                                                                                          
      × should be chunked after concurrent with filter (1055 ms)  // around 1050ms                                                                                                                                
[...]

PR #137, opened 2022-02-03
feat: add `fromEntries` function

Kuhave:
Hi all,
I wrote a `fromEntries` function that performs inverse of `entries` function(key-value pair iterable/asyncIterable to an object).
This function supports type inference well.

My English is not very good, so there may be some strange English descriptions.

And also, maybe my test cases are weird or poorly written.

Any advice would be welcome. 😄

### DIFF
--- a/src/fromEntries.ts
+++ b/src/fromEntries.ts
@@ -1,0 +1,67 @@
+import IterableInfer from "./types/IterableInfer";
+import ReturnValueType from "./types/ReturnValueType";
+import { isAsyncIterable, isIterable } from "./_internal/utils";
+
+async function async<T extends [string, any]>(
+  iterable: AsyncIterable<T>,
+): Promise<{
+  [K in T as K[0]]: K[1];
+}> {
+  const a = {} as any;
+  for await (const i of iterable) {
+    a[i[0]] = i[1];
+  }
+  return a;
+}
+
+/**
+ * Returns an object from string keyed-value pairs.
+ *
+ * @example
+ * ```ts
+ * const arr = [
+ *     ["a", 1],
+ *     ["b", true],
+ *     ["c", "hello"],
+ *     ["d", { d1: 1, d2: 3 }],
+ * ] as (
+ *     | ["a", number]
+ *     | ["b", boolean]
+ *     | ["c", string]
+ *     | ["d", { d1: number, d2: number; }]
+ * )[];
+ * fromEntries(arr); // { a: 1, b: true, c: 'hello', d: { d1: 1, d2: 3 } }
+ * ```
+ *
+ *
+ * see {@link https://fxts.dev/docs/entries | entries}
+ */
+
+function fromEntries<
+  U extends [string, any] | readonly [string, any],
+  T extends Iterable<U> | AsyncIterable<U>,
+>(
+  iterable: T,
+): ReturnValueType<
+  T,
+  {
+    [K in IterableInfer<T> as K[0]]: K[1];
+  }
+>;
+function fromEntries<T extends [string, any]>(
+  iter: Iterable<T> | AsyncIterable<T>,
+) {
+  if (isAsyncIterable(iter)) {
+    return async(iter);
+  } else if (isIterable(iter)) {
+    return [...iter].reduce((obj, [key, val]) => {
+      obj[key] = val;
+      return obj;
+    }, {} as any);
+  }
+  return {} as {
+    [K in T as K[0]]: K[1];
+  };
+}
+
+export default fromEntries;
--- a/src/fromEntries.ts
+++ b/src/fromEntries.ts
@@ -1,8 +1,9 @@
 import IterableInfer from "./types/IterableInfer";
+import Key from "./types/Key";
 import ReturnValueType from "./types/ReturnValueType";
 import { isAsyncIterable, isIterable } from "./_internal/utils";
 
-async function async<T extends [string, any]>(
+async function async<T extends [Key, any]>(
   iterable: AsyncIterable<T>,
 ): Promise<{
   [K in T as K[0]]: K[1];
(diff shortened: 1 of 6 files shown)


hg-pyun:
This code failed to pass the test case. Please check it.

test/fromEntries.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+        | ["a", number]
+        | ["b", true]
+        | ["c", string]
+        | ["d", { d1: number; d2: number }]
+      )[];
+      const res = fromEntries(arr);
+      expect(res).toEqual({ a: 1, b: true, c: "hello", d: { d1: 1, d2: 3 } });
+    });
+  });
+});
```
hg-pyun, line 22: It would be nice to add a test case for [ES6 Map](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#map%EC%97%90%EC%84%9C_object%EB%A1%9C).

Kuhave:
> This code failed to pass the test case. Please check it.

@hg-pyun I added a few more test cases as per your advice 😄 

In my opinion, the reason the test doesn't pass seems to be related to this existing issue #93 .

I confirmed that the test cases in this PR passed.

image

test/fromEntries.spec.ts
Kuhave: I added a few more test cases. thanks!

hg-pyun:
@kuhave Oh, I see. The final approval is delegated to @shine1594 and @ppeeou

hg-pyun (approved):

ppeeou:
Thank you for your contribution👍
I left a few comments

src/fromEntries.ts
```diff
@@ -0,0 +1,68 @@
...
+    [K in IterableInfer<T> as K[0]]: K[1];
+  }
+>;
+function fromEntries<T extends [Key, any]>(
+  iter: Iterable<T> | AsyncIterable<T>,
+) {
+  if (isAsyncIterable(iter)) {
+    return async(iter);
+  } else if (isIterable(iter)) {
+    return [...iter].reduce((obj, [key, val]) => {
```
ppeeou, line 58: How about checking if it is an array and doing `iter.reduce(...)` without making a thin copy of the array?

src/fromEntries.ts
```diff
@@ -0,0 +1,68 @@
...
+      obj[key] = val;
+      return obj;
+    }, {} as any);
+  }
+  return {} as {
+    [K in T as K[0]]: K[1];
+  };
```
ppeeou, line 65: How about raising a TypeError because the user entered an invalid argument value?

src/fromEntries.ts
```diff
@@ -0,0 +1,68 @@
...
+  iter: Iterable<T> | AsyncIterable<T>,
+) {
+  if (isAsyncIterable(iter)) {
+    return async(iter);
+  } else if (isIterable(iter)) {
+    return [...iter].reduce((obj, [key, val]) => {
+      obj[key] = val;
+      return obj;
+    }, {} as any);
+  }
```
shine1594, line 62: If you use the reduce function of `fxts`, you do not need to write a separate `async` function.

```suggestion
const obj: Record<string, any> = {};
  if (isAsyncIterable(iter)) {
    return reduce(
      (obj, [key, val]) => {
        obj[key] = val;
        return obj;
      },
      obj,
      iter,
    ) as any;
  } else if (isIterable(iter)) {
    return reduce(
      (obj, [key, val]) => {
        obj[key] = val;
        return obj;
      },
      obj,
      iter,
    ) as any;
  }
```

src/fromEntries.ts
shine1594: I agree because an error occurs if an argument other than `iterable` is given to the `Object.fromEntries` method.
스크린샷 2022-02-06 오후 11 09 47

type-check/fromEntries.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const res1 = fromEntries(arr);
+const res2 = pipe(obj, entries, fromEntries);
+const res3 = pipe(obj, entries, toAsync, fromEntries);
+
+checks([
+  check<typeof res1, typeof obj, Test.Pass>(),
+  check<typeof res2, typeof obj, Test.Pass>(),
+  check<typeof res3, Promise<typeof obj>, Test.Pass>(),
+]);
```
shine1594, line 33: Since the types of `res2` and `res3` may be inferred differently depending on implementation, it would be better to add a test case like this.
```suggestion
const res1 = fromEntries(arr);
const res2 = pipe(obj, entries, fromEntries);
const res3 = pipe(obj, entries, (iter)  => fromEntries(iter));
const res4 = pipe(obj, entries, toAsync, fromEntries);
const res5 = pipe(obj, entries, toAsync, (iter) => fromEntries(iter));

checks([
  check<typeof res1, typeof obj, Test.Pass>(),
  check<typeof res2, typeof obj, Test.Pass>(),
  check<typeof res3, typeof obj, Test.Pass>(),
  check<typeof res4, Promise<typeof obj>, Test.Pass>(),
  check<typeof res5, Promise<typeof obj>, Test.Pass>(),
]);
```

shine1594 (changes requested):

src/fromEntries.ts
Kuhave: yeah I also agree with you.
I'll change it with throws a TypeError here. 👍

type-check/fromEntries.test.ts
Kuhave: thanks. i'll change with your code. 👍

src/fromEntries.ts
Kuhave: i applied it. thanks!

shine1594 (approved):

ppeeou (approved):
LGTM 👍

PR #140, opened 2022-02-07
fix: fix return type in Curry type

### LINKED ISSUE #139: The curried function returned from the fxts/curry is incorrectly inferred.
# Bug Report

### 💻 Code

```ts
import { each, map, pipe, curry } from '@fxts/core';

const add = curry((a: number, b: number) => a + b);

pipe([1, 2, 3], map(add(10)), map(add(100)), each(console.log));
```

### 🙁 Actual behavior

Typescript can't correctly infer the type of add when it executed in map or pipe.

* add(10) in first map
image

* add(100) in second map
image

More precisely, every functions whose type is compatible with `<X>(f: (...args) => X): X` can't infer if callback is executed curried function.

`Curry<F extends (...args: any[]) => any>` type return the below type.
```ts
<
  T extends any[],
  G = Drop<Length<T>, Parameters<F>>,
  R = ReturnType<F>,
>(
  ...args: Cast<T, Gaps<Parameters<F>>>
) => G extends [any, ...any[]] ? Curry<(...args: G) => R> : R;
```

I think typescript infers type `f: (...args) => X`, not type `X` as the return type of curried function, because `ReturnType` uses the `infer` keyword and Curry passes `R = ReturnType` recursively.

So we need to ensure that the type of `ReturnType<F>` can be fixed only once when the `curry` function is initially executed.
```ts
<
  T extends any[],
  G = Drop<Length<T>, Parameters<F>>,
>(
  ...args: Cast<T, Gaps<Parameters<F>>>
) => G extends [any, ...any[]]
  ? Curry<(...args: G) => ReturnType<F>>
  : ReturnType<F>;
```

### 🙂 Expected behavior

* add(10) in first map
image

* add(100) in second map
image

---

### Version Information

- Operating system: macOS Big Sur 11.1
- Typescript: ^4.5.4
- Nodejs: v17.3.1

degurii:
I wrote the details on the #139 

Fixes #139

### DIFF
--- a/type-check/curry.test.ts
+++ b/type-check/curry.test.ts
@@ -40,5 +40,5 @@ checks([
   check<typeof res6, AsyncIterableIterator<number>, Test.Pass>(),
   check<typeof res7, number, Test.Pass>(),
   check<typeof res8, IterableIterator<number>, Test.Pass>(),
-  check<typeof res9, AsyncIterableIterator<number>, Test.Fail>(),
+  check<typeof res9, AsyncIterableIterator<number>, Test.Pass>(),
 ]);
--- a/type-check/curry.test.ts
+++ b/type-check/curry.test.ts
@@ -1,7 +1,7 @@
 import * as Test from "../src/types/Test";
 import Curry from "../src/types/Curry";
 import { curry, pipe, toAsync } from "../src";
-import { filter } from "../src/Lazy";
+import { filter, map } from "../src/Lazy";
 
 const { checks, check } = Test;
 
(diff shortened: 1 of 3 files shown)


ppeeou (commented):
thank you for your contribution 👍

Can you add type result test in `curry.test.ts` as well?

type-check/curry.test.ts
```diff
@@ -34,4 +38,7 @@ checks([
   check<typeof res4, string, Test.Pass>(),
   check<typeof res5, IterableIterator<number>, Test.Pass>(),
   check<typeof res6, AsyncIterableIterator<number>, Test.Pass>(),
+  check<typeof res7, number, Test.Pass>(),
+  check<typeof res8, IterableIterator<number>, Test.Pass>(),
+  check<typeof res9, AsyncIterableIterator<number>, Test.Fail>(),
```
ppeeou, line 43:
```suggestion
  check<typeof res9, AsyncIterableIterator<number>, Test.Pass>(),
```

degurii:
Oh, I made a typo...  res9 should be Test.Pass() 😭

ppeeou (approved):
LGTM👍

PR #141, opened 2022-02-10
fix: replace duplicated Iterable type with UniversalIterable

lifeisegg123:
I was wondering why you didn't use `UniversalIterable` and `UniversalIterator`.
If there is reason to avoid using it, it would be lovely to explain me why 🙏


shine1594:
Of course, there are no technical problems 😄  However, I think it is better to use the built-in type as it is for users who read the document. I don't want them to spend time looking up what `UniversialIterable` and `UniversialIterator` are.

PR #144, opened 2022-02-15
feat: add `takeRight` function

lifeisegg123:
Hi!
Suggestion with PR. I saw that there's a `dropRight` function but not `takeRight`.

As far as I know, some of the fp libraries (lodash, ramda) has this function, and I think it's good to have one in FxTS as well.

I don't think my codes works 100% perfect, maybe it's better rewrite the whole code by maintainers.
But it would be nice if you could give me any advice, and more it would be appreciated if this PR accepted.

Thanks for reading my codes in advance, sorry for my poor English and coding skill 🙏
I hope this library getting more famous and adopted over the others ❤️

### DIFF
--- a/src/Lazy/takeRight.ts
+++ b/src/Lazy/takeRight.ts
@@ -9,8 +9,8 @@ import { isConcurrent } from "./concurrent";
 
 function* stringTakeRight(length: number, str: string) {
   const arr = Array.from(str);
-  const index = str.length - length;
-  for (let i = index; i < str.length; i++) {
+  const index = arr.length - length;
+  for (let i = index; i < arr.length; i++) {
     if (arr[i]) yield arr[i];
   }
 }
--- a/src/Lazy/takeRight.ts
+++ b/src/Lazy/takeRight.ts
@@ -1,0 +1,149 @@
+import { isAsyncIterable, isIterable } from "../_internal/utils";
+import ReturnIterableIteratorType from "../types/ReturnIterableIteratorType";
+import IterableInfer from "../types/IterableInfer";
+import toArray from "../toArray";
+import isArray from "../isArray";
+import isString from "../isString";
+import concurrent from "./concurrent";
+import { isConcurrent } from "./concurrent";
+
+function* stringTakeRight(length: number, str: string) {
+  const arr = Array.from(str);
+  const index = str.length - length;
+  for (let i = index; i < str.length; i++) {
+    if (arr[i]) yield arr[i];
+  }
+}
+
+function* arrayTakeRight<A>(length: number, arr: Array<A>) {
+  const index = arr.length - length;
+  for (let i = index; i < arr.length; i++) {
+    if (arr[i]) yield arr[i];
+  }
+}
+
+function* sync<A>(length: number, iterable: Iterable<A>): IterableIterator<A> {
+  const arr = toArray(iterable);
+  const index = arr.length - length;
+  for (let i = index; i < arr.length; i++) {
+    if (arr[i]) yield arr[i];
+  }
+}
+
+async function* asyncSequential<T>(length: number, iterable: AsyncIterable<T>) {
+  const arr = await toArray(iterable);
+  const index = arr.length - length;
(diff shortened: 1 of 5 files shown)


ppeeou (commented):
Thank you for your contribution:)👍
 
I left a few comments

src/Lazy/takeRight.ts
```diff
@@ -0,0 +1,149 @@
...
+):
+  | IterableIterator<IterableInfer<A>>
+  | AsyncIterableIterator<IterableInfer<A>>
+  | ((iterable: A) => ReturnIterableIteratorType<A>) {
+  if (iterable === undefined) {
+    return (iterable: A) => {
+      return takeRight(l, iterable as any) as ReturnIterableIteratorType<A>;
+    };
+  }
+
```
ppeeou, line 129: Since there is a logic that evaluates everything with toArray, it seems to be necessary to check the correct argument.

```ts
  if (length < 1) {
    throw new TypeError("'length' must be greater than 1");
  }
```

test/Lazy/takeRight.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        toAsync([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
+        map((a) => delay(100, a)),
+        filter((a) => a % 2 === 0),
+        takeRight(3),
+        concurrent(3),
+        toArray,
+      );
+
+      expect(res).toEqual([6, 8, 10]);
+    }, 450);
```
ppeeou, line 116: Concurrent evaluation should also add tests that only evaluate exactly a given number. (3 in the test above)

```suggestion
    it("should be able to take the element concurrently", async function () {
      const fn = jest.fn();
      callFuncAfterTime(fn, 400);
      const res = await pipe(
        toAsync([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
        map((a) => delay(100, a)),
        filter((a) => a % 2 === 0),
        takeRight(3),
        concurrent(3),
        toArray,
      );
      expect(fn).toBeCalled();
      expect(res).toEqual([6, 8, 10]);
    }, 450);
```

src/Lazy/takeRight.ts
lifeisegg123: Thanks for your kind comment!
I added this logic and please check it if I'm correct!

test/Lazy/takeRight.spec.ts
lifeisegg123: this too!

ppeeou (approved):
lgtm👍

test/Lazy/takeRight.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  range,
+  takeRight,
+  toArray,
+  toAsync,
+  concurrent,
+} from "../../src/index";
+import { Concurrent } from "../../src/Lazy/concurrent";
+import { callFuncAfterTime, generatorMock } from "../utils";
+
+describe("take", function () {
```
shine1594, line 15:
```suggestion
describe("takeRight", function () {
```

src/Lazy/takeRight.ts
```diff
@@ -0,0 +1,152 @@
...
+import toArray from "../toArray";
+import isArray from "../isArray";
+import isString from "../isString";
+import concurrent from "./concurrent";
+import { isConcurrent } from "./concurrent";
+
+function* stringTakeRight(length: number, str: string) {
+  const arr = Array.from(str);
+  const index = str.length - length;
+  for (let i = index; i < str.length; i++) {
```
shine1594, line 13:
```suggestion
  const index = arr.length - length;
  for (let i = index; i < arr.length; i++) {
```

shine1594 (changes requested):

test/Lazy/takeRight.spec.ts
lifeisegg123: thanks for your suggestion and I changed this and below one!

shine1594 (approved):
LGTM

PR #146, opened 2022-03-10
fix: minimum length to be possible from 0

shine1594:


hg-pyun (approved):
LGTM, It would be nice to explain why the PR content changed to zero.

ppeeou (approved):
lgtm

shine1594:
@hg-pyun 
While we won't explicitly use length(or count) as zero, this is useful because it can be zero if computed dynamically 🙂

Issue #148, opened 2022-03-31
Support to handle unicode


ppeeou:
# Suggestion

## ⭐ Suggestion

now fxts can not handle unicode
so suggest to `unicodeToArray ` like https://github.com/lodash/lodash/blob/master/.internal/unicodeToArray.js

## 💻 Use Cases

```ts
pipe(
  "🙇‍♂️🌈⛅️🔥",
   unicodeToArray,
   reverse,
   ...
)

```

PR #149, opened 2022-03-31
Fix/unicode

shine1594:
Fixes https://github.com/marpple/FxTS/issues/148

Instead of handling unicode with the `Array.from` function in dropRight (and reverse, takeRight), the `unicodeToArray` function of `lodash.js` is provided, so that the user can directly process the unicode as needed.

### DIFF
--- a/src/unicodeToArray.ts
+++ b/src/unicodeToArray.ts
@@ -1,0 +1,58 @@
+// source: https://github.com/lodash/lodash/blob/master/.internal/unicodeToArray.js
+
+/** Used to compose unicode character classes. */
+const rsAstralRange = "\\ud800-\\udfff";
+const rsComboMarksRange = "\\u0300-\\u036f";
+const reComboHalfMarksRange = "\\ufe20-\\ufe2f";
+const rsComboSymbolsRange = "\\u20d0-\\u20ff";
+const rsComboMarksExtendedRange = "\\u1ab0-\\u1aff";
+const rsComboMarksSupplementRange = "\\u1dc0-\\u1dff";
+const rsComboRange =
+  rsComboMarksRange +
+  reComboHalfMarksRange +
+  rsComboSymbolsRange +
+  rsComboMarksExtendedRange +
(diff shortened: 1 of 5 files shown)


src/unicodeToArray.ts
```diff
@@ -0,0 +1,58 @@
...
+const rsSymbol = `(?:${[
+  rsNonAstralCombo,
+  rsCombo,
+  rsRegional,
+  rsSurrPair,
+  rsAstral,
+].join("|")})`;
+
+/** Used to match [string symbols](https://mathiasbynens.be/notes/javascript-unicode). */
+const reUnicode = RegExp(`${rsFitz}(?=${rsFitz})|${rsSymbol + rsSeq}`, "g");
```
hg-pyun, line 45: Incredible! I used to work on something similar, but lodash made it really well. 👍

hg-pyun (approved):

hg-pyun:
@shine1594 
I don't know if FxTS should provide this function. I think I can definitely use it from the outside...

ppeeou:
> @shine1594 Additionally, I don't know if FxTS should provide this function. I think I can definitely use it from the outside...

@hg-pyun 
Do you mean by using  `unicodeToArray` if a `string` value is entered in an operation like `toArray`? without exposing the function?

hg-pyun:
@ppeeou It meant that the user could define and import it himself.

src/unicodeToArray.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Converts a Unicode `string` to an array.
+ *
+ * @private
+ * @param {string} string The string to convert.
+ * @returns {Array} Returns the converted array.
```
ppeeou, line 52:
```suggestion
* @example
* ```ts
*  unicodeToArray('🙇‍♂️🤩😭'); // ['🙇‍♂️','🤩','😭'];
* 
* ```
```

ppeeou (approved):

PR #152, opened 2022-04-08
feat: lt, gt, lte, gte

### LINKED ISSUE #151: lt, lte, gt, gte function
# Suggestion

## ⭐ Suggestion

Provides a size comparison function.

It does not provide a generic type, so we are going to provide curry by default.

## 💻 Use Cases

```ts
function lte(a:number, b:number) {
  return a <= b
}

function lt(a:number, b:number) {
  return a < b
}

function gte(a:number, b:number) {
  return a >= b
}

function gt(a:number, b:number) {
  return a > b
}
```

sunrabbit123:
Fixes #151 

But I'm not sure if this is what you want from the issue.

### DIFF
--- a/src/gt.ts
+++ b/src/gt.ts
@@ -24,7 +24,7 @@ function gt(a: Date, b: Date): boolean;
 
 function gt(a: any, b?: any): ((b: any) => boolean) | boolean {
   if (b === undefined) {
-    return (_b: any) => gt(_b, a);
+    return (_b: any) => gt(a, _b);
   }
 
   if (a.constructor !== b.constructor) {
--- a/src/gt.ts
+++ b/src/gt.ts
@@ -1,6 +1,3 @@
-import curry from "./curry";
-import Curry from "./types/Curry";
-
 /**
  * Returns true if the first argument is greater than the second; false otherwise.
  *
@@ -17,17 +14,24 @@ import Curry from "./types/Curry";
  * filter(gt("b"), ["a", "b"]) // Itreable<[]>
  * ```
  */
-function gt(a: string): Curry<(b: string) => boolean>;
-function gt(a: number): Curry<(b: number) => boolean>;
+
+function gt(a: string | number): (b: string | number) => boolean;
 function gt(a: string, b: string): boolean;
 function gt(a: number, b: number): boolean;
 
 function gt(
   a: string | number,
   _b?: string | number,
-): Curry<(b: string | number) => boolean> | boolean {
+): ((b: string | number) => boolean) | boolean {
   if (_b === undefined) {
-    return curry((c: string | number, b: string | number): boolean => b > c)(a);
+    return (b: string | number): boolean => {
+      if (typeof a !== typeof b) {
+        throw new TypeError(
+          "The values you want to compare must be of the same type",
+        );
+      }
+      return b > a;
+    };
   }
 
   return a > _b;
--- a/src/gt.ts
+++ b/src/gt.ts
@@ -2,11 +2,16 @@ import curry from "./curry";
 import Curry from "./types/Curry";
 
 /**
- * Make sure that it is less than the reference value.
+ * Make sure that it is greater than the reference value.
  * It does not provide a generic type, so we are going to provide curry by default.
  *
  * @example
  * ```ts
+ * gt(5, 1) // expected true
+ * gt(1, 5) // expected false
+ * gt("a", "b") // expected false
+ * gt("b", "a") // expected true
+ *
  * filter(gt(5), [1, 2, 4, 5, 8, 9]) // Iterable<[8, 9]>
  * filter(gt(5), [1, 2, 3, 4, 5]) // Iterable<[]>
  * filter(gt("b"), ["a", "b", "c"]) // Iterable<["c"]>
@@ -15,9 +20,18 @@ import Curry from "./types/Curry";
  */
 function gt(a: string): Curry<(b: string) => boolean>;
 function gt(a: number): Curry<(b: number) => boolean>;
+function gt(a: string, b: string): boolean;
+function gt(a: number, b: number): boolean;
 
-function gt(a: string | number): Curry<(b: string | number) => boolean> {
-  return curry((c: string | number, b: string | number): boolean => b > c)(a);
+function gt(
+  a: string | number,
+  _b?: string | number,
+): Curry<(b: string | number) => boolean> | boolean {
+  if (_b === undefined) {
+    return curry((c: string | number, b: string | number): boolean => b > c)(a);
+  }
+
+  return a > _b;
 }
 
 export default gt;
--- a/src/gt.ts
+++ b/src/gt.ts
@@ -15,27 +15,25 @@
  * ```
  */
 
-function gt(a: string | number | Date): (b: string | number | Date) => boolean;
+function gt(a: string): (b: string) => boolean;
+function gt(a: number): (b: number) => boolean;
+function gt(a: Date): (b: Date) => boolean;
 function gt(a: string, b: string): boolean;
 function gt(a: number, b: number): boolean;
 function gt(a: Date, b: Date): boolean;
 
-function gt(
-  a: string | number | Date,
-  _b?: string | number | Date,
-): ((b: string | number | Date) => boolean) | boolean {
-  if (_b !== undefined) {
-    return a > _b;
+function gt(a: any, b?: any): ((b: any) => boolean) | boolean {
+  if (b === undefined) {
+    return (_b: any) => gt(_b, a);
   }
 
-  return (b: string | number | Date): boolean => {
-    if (typeof a !== typeof b) {
-      throw new TypeError(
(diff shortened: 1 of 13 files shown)


src/gt.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @example
+ * ```ts
+ * filter(gt(5), [1, 2, 4, 5, 8, 9]) // Iterable<[8, 9]>
+ * filter(gt(5), [1, 2, 3, 4, 5]) // Iterable<[]>
+ * filter(gt("b"), ["a", "b", "c"]) // Iterable<["c"]>
+ * filter(gt("b"), ["a", "b"]) // Itreable<[]>
+ * ```
+ */
+function gt(a: string): Curry<(b: string) => boolean>;
+function gt(a: number): Curry<(b: number) => boolean>;
```
ppeeou, line 17: I think the signature below should also be provided.
```ts
function gt(a: number, b:number): boolean;
```

ppeeou (commented):
Thank you for your contribution👍

I left a few comments

src/gt.ts
sunrabbit123: I don't understand

I understood this.
Is this right?

```ts
function gt(a: number, b: number) : boolean { return a > b; }

const result = gt(5, 1); // expected true
```

ppeeou (approved):
It's the same as what I intended
https://github.com/marpple/FxTS/pull/152/commits/475efe29ecc4fee793a28a46328c9e2d92afbea9 👍

test/gt.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+      expect(result).toEqual(["d", "e"]);
+    });
+    it("given string array then should return empty array", function () {
+      const result = pipe(["a", "b", "c", "d"], filter(gt("d")), toArray);
+      expect(result).toEqual([]);
+    });
+  });
+
+  describe("gt given boolean", function () {
+    it("given two number should return true", function () {
```
ppeeou, line 25: How about writing the testcase in the following pattern?
`should return true that first argument is greater than second argument`

src/gt.ts
```diff
@@ -0,0 +1,37 @@
+import curry from "./curry";
+import Curry from "./types/Curry";
+
+/**
+ * Make sure that it is greater than the reference value.
+ * It does not provide a generic type, so we are going to provide curry by default.
```
ppeeou, line 6:
```suggestion
`Returns true if the first argument is greater than the second; false otherwise.`
```

`curry` is confirmed in the signature, so there is no need to write it separately.

sunrabbit123: It is good suggestion, thx :D 

Even I think it's better to avoid duplicates.

test/gt.spec.ts
sunrabbit123: I didn't know that.


Reference was made to the test case name of 'isArray.spec.ts'.
However, the other test case names are different.

Now that I know it, I will revise it.

thx, :D

ppeeou:
@shine1594 @hg-pyun  Could anyone please check this PR?

src/gte.ts
```diff
@@ -0,0 +1,39 @@
...
+ * gte("a", "b") // expected false
+ * gte("b", "a") // expected true
+ *
+ * filter(gte(5), [1, 2, 4, 5, 8, 9]) // Iterable<[5, 8, 9]>
+ * filter(gte(5), [1, 2, 3, 4]) // Iterable<[]>
+ * filter(gte("b"), ["a", "b", "c"]) // Iterable<["b", "c"]>
+ * filter(gte("b"), ["a"]) // Itreable<[]>
+ * ```
+ */
+function gte(a: string): Curry<(b: string) => boolean>;
```
shine1594, line 21: It should be able to accept the `Date` object as an argument as well.

sunrabbit123: It's good idea, I will develop this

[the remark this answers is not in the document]
sunrabbit123: I think that this is good opinion, I think I focused on Curry.

sunrabbit123: but I have a question

compare date? or compare time?
The standards are ambiguous

shine1594: I mean just like this:

```ts
function gte(a: Date, b: Date) {
  return a >= b;
}

const d1 = new Date();
const d2 = new Date(d1);

gte(d1, d2); // true
```

sunrabbit123: my question mean like this:
```ts
function gte(a : Date, b : Date){
  return a > b;
}

const d1 = new Date(2022, 4, 11);
const d2 = new Date(2022, 4, 11, 4, 10 ,10);

gte(d1, d2); // true? or false?
```

struggle between plan a that compare `a < b` and plan b that compare `a.getTime() < b.getTime()`

Maybe like this...
```ts
const compare_a = a instanceof Date ? a.getTime() : a;
return (b: string | number | Date): boolean => {
    if (typeof a !== typeof b) {
      throw new TypeError(
        "The values you want to compare must be of the same type",
      );
    }
    const compare_b = b instanceof Date ? b.getTime() : b;
    return compare_b < compare_a;
  };
```

I don't know the universal plan.

shine1594: There is no reason to use the `getTime` method. Just using the comparison operator is enough.

sunrabbit123: Um... I think some user could use this to compare time.

But for now, I made it what you wanted.

sunrabbit123: !.!
I realized this example is wrong
`gt(d1, d2); // true? or false?` 

I'm so sorry

thx.

shine1594 (changes requested):

src/gt.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  return (b: string | number | Date): boolean => {
+    if (typeof a !== typeof b) {
+      throw new TypeError(
+        "The values you want to compare must be of the same type",
+      );
+    }
+    return b > a;
+  };
+}
```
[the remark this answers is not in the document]
sunrabbit123: 'gt' of 'lodash' does not provide 'curry'.
`ramda`... is I don't know

The reason why the order has changed is because of `curry`.

The test code is as follows:
```ts
it("given array then should return greater values", function () {
      const result = pipe([4, 5, 6], filter(gt(5)), toArray);
      expect(result).toEqual([6]);
});
```

it is faild at `gt(a, _b)` but `gt(_b, a)` is success

```ts
function gt(a: any, b?: any): ((b: any) => boolean) | boolean {
  if (b === undefined) {
    return (_b: any) => gt(_b, a);
  }

  if (a.constructor !== b.constructor) {
    throw new TypeError(
      "The values you want to compare must be of the same type",
    );
  }

  return a > b;
}
```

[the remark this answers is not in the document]
shine1594: ### A
```ts
pipe([4, 5, 6], filter(gt(5)), toArray); // [6]
```

### B
```ts
pipe([4, 5, 6], filter(lt(5)), toArray); // [6]
```

I agree that A looks more natural than B. However, many users will think that `gt(5)` is a curryed form of `gt(5, x)`, so it is easy to predict that it will evaluate to `5 > x`.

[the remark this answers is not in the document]
sunrabbit123: It's vague, but I understand.

I will work to change test code

src/lt.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Returns true if the first argument is less than the second; false otherwise.
+ *
+ * @example
+ * ```ts
+ * lte(5, 1) // expected false
+ * lte(1, 5) // expected true
+ * lte("a", "b") // expected true
+ * lte("b", "a") // expected false
```
shine1594, line 9:
```suggestion
 * lt(5, 1) // expected false
 * lt(1, 5) // expected true
 * lt("a", "b") // expected true
 * lt("b", "a") // expected false
```

src/lt.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * filter(lt(5), [1, 2, 4, 5, 8, 9]) // Iterable<[5, 8, 9]>
```
shine1594, line 11:
```suggestion
 * filter(lt(5), [1, 2, 4, 5, 8, 9]) // Iterable<[8, 9]>
```

test/lt.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { filter, lt, pipe, toArray } from "../src/index";
+
+describe("lt(little)", function () {
```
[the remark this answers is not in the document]
sunrabbit123: I fault, I'm embarrassed

src/lt.ts
```diff
@@ -0,0 +1,38 @@
...
+ * filter(lt(7), [5, 6, 7]) // Iterable<[]>
+ * filter(lt("b"), ["a", "b", "c"]) // Iterable<["b", "c"]>
```
shine1594, line 13:
```suggestion
 * filter(lt("b"), ["a", "b", "c"]) // Iterable<["c"]>
```

PR #162, opened 2022-05-25
feat: implement `compactObject`

shine1594:
Add `compactObject` function.


src/compactObject.ts
```diff
@@ -0,0 +1,47 @@
...
+    [K, any]
+  >[1];
+};
+
+/**
+ * Returns an object with all nullable values removed.
+ *
+ * @example
+ * ```ts
+ * const compacted = compact({ a: 1, b:"b", c: null, d: undefined });
```
ppeeou, line 37:
```suggestion
 * const compacted = compactObject({ a: 1, b:"b", c: null, d: undefined });
```

ppeeou (approved):
lgtm

PR #164, opened 2022-05-31
feat: support thenable

### LINKED ISSUE #163: Support `thenable` object  to be used like `Promise`
# Suggestion

## ⭐ Suggestion

The thenable object is often also used like a promise object.

fxts is a functional library, and its combination with other libraries is important.

When returning an asynchronous function from another library, there are cases where it does not return a Promise object but returns it according to the Promsie spec.

Below is an example of prisma.

https://github.com/prisma/prisma/blob/3855438b964b40edeb72a3815c641d97f8c433c4/packages/client/src/runtime/core/request/createPrismaPromise.ts#L15-L55

```ts
export function createPrismaPromise(
  callback: (txId?: string | number, lock?: PromiseLike<void>, otelCtx?: Context) => PrismaPromise<unknown>,
): PrismaPromise<unknown> {
  ...
  return {
    then(onFulfilled, onRejected, txId?: string) {
      return _callback(txId).then(onFulfilled, onRejected, txId)
    },
    catch(onRejected, txId?: string) {
      return _callback(txId).catch(onRejected, txId)
    },
    finally(onFinally, txId?: string) {
      return _callback(txId).finally(onFinally, txId)
    },
   ...
```

So I think fxts should also support theanble object.

ppeeou:
Fixes #163

### DIFF
--- a/src/_internal/utils.ts
+++ b/src/_internal/utils.ts
@@ -40,3 +40,20 @@ export const asyncEmpty = async function* () {};
 
 export const isNotNullable = <T>(a: T): a is NonNullable<T> =>
   a !== null && a !== undefined;
+
+export const isPromise = <T>(a: Promise<T> | T): a is Promise<T> => {
+  if (a instanceof Promise) {
+    return true;
+  }
+
+  if (
+    a !== null &&
+    typeof a === "object" &&
+    typeof (a as any).then === "function" &&
+    typeof (a as any).catch === "function"
+  ) {
+    return true;
+  }
+
+  return false;
+};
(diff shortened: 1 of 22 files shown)


src/_internal/utils.ts
```diff
@@ -40,3 +41,20 @@ export const asyncEmpty = async function* () {};
...
+export const isPromise = <T>(a: Promise<T> | T): a is Promise<T> => {
+  if (a instanceof Promise) {
+    return true;
+  }
+
+  if (
+    a !== null &&
+    typeof a === "object" &&
+    typeof (a as any).then === "function" &&
+    typeof (a as any).catch === "function"
```
ppeeou, line 54: It seems to be necessary to discuss at what level the judgment of the thenable object should be decided.

In the case of co, only the existence of the then function is judged.
https://github.com/tj/co/blob/master/index.js#L198

shine1594: I also think this level would be appropriate.

src/_internal/utils.ts
```diff
@@ -1,3 +1,4 @@
+import isNil from "../isNil";
```
shine1594, line 1:
```suggestion
```

shine1594 (approved):
LGTM

PR #169, opened 2022-08-13
fix(sum): sum function logic

### LINKED ISSUE #156: The empty array result of the `sum` function is '0'
# Suggestion

## ⭐ Suggestion

If an empty array is entered as the argument value of the [sum function](https://fxts.dev/docs/sum), undefined is output.
Returning 0 seems more natural.

as-is
```ts
const result = pipe(
 [], 
 sum
);

// result undefined
```

to-be
```ts
const result = pipe(
 [], 
 sum
);

// result 0
```

hg-pyun:
- only supoort number type
- empty array returns zero

Fixes #156

### DIFF
--- a/src/sum.ts
+++ b/src/sum.ts
@@ -17,11 +17,14 @@ function sum<A extends Iterable<number> | AsyncIterable<number>>(
   ? number
   : A extends AsyncIterable<number>
   ? Promise<number>
-  : never {
+  : never;
+function sum<A extends Iterable<number> | AsyncIterable<number>>(
+  iterable: A,
+): number | Promise<number> {
   if (Array.isArray(iterable)) {
     return iterable.reduce(add, 0);
   } else if (isIterable(iterable)) {
-    return reduce(add, 0, iterable);
+    return reduce<number, number>((a, b) => add(a, b), 0, iterable);
   } else if (isAsyncIterable(iterable)) {
     return reduce<number, number>(add, Promise.resolve(0), iterable);
   }
--- a/src/sum.ts
+++ b/src/sum.ts
@@ -1,6 +1,5 @@
 import add from "./add";
 import reduce from "./reduce";
-import ReturnValueType from "./types/ReturnValueType";
 import { isAsyncIterable, isIterable } from "./_internal/utils";
 
 /**
@@ -12,18 +11,17 @@ import { isAsyncIterable, isIterable } from "./_internal/utils";
  * await sum(toAsync([1, 2, 3, 4])); // 10
  * ```
  */
-function sum<A extends readonly []>(iterable: A): 0;
-function sum<A extends number[]>(arr: A): number;
-function sum<A extends Iterable<number>>(iterable: A): number;
-function sum<A extends AsyncIterable<number>>(iterable: A): Promise<number>;
-
 function sum<A extends Iterable<number> | AsyncIterable<number>>(
   iterable: A,
-): number | Promise<number> {
+): A extends Iterable<number>
+  ? number
+  : A extends AsyncIterable<number>
+  ? Promise<number>
+  : never {
   if (Array.isArray(iterable)) {
     return iterable.reduce(add, 0);
   } else if (isIterable(iterable)) {
-    return reduce(add, iterable) as ReturnValueType<A>;
+    return reduce(add, 0, iterable);
   } else if (isAsyncIterable(iterable)) {
     return reduce<number, number>(add, Promise.resolve(0), iterable);
   }
--- a/src/sum.ts
+++ b/src/sum.ts
@@ -14,19 +14,33 @@ import { isAsyncIterable, isIterable } from "./_internal/utils";
  * await sum(toAsync(['a', 'b', 'c'])); // 'abc'
  * ```
  */
-function sum<
-  A extends
-    | Iterable<number>
-    | AsyncIterable<number>
-    | Iterable<string>
-    | AsyncIterable<string>,
->(iterable: A): ReturnValueType<A> {
+function sum<A extends readonly []>(iterable: A): 0;
+function sum<A extends Iterable<number> | AsyncIterable<number>>(
+  iterable: A,
+): ReturnValueType<A>;
+
+function sum<A extends Iterable<number> | AsyncIterable<number> | readonly []>(
+  iterable: A,
+): ReturnValueType<A> | 0 {
   if (isIterable(iterable)) {
+    const iterator = iterable[Symbol.iterator]();
+    const { done } = iterator.next();
+    if (done) {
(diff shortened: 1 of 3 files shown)


src/sum.ts
```diff
@@ -14,19 +14,33 @@ import { isAsyncIterable, isIterable } from "./_internal/utils";
...
   if (isAsyncIterable(iterable)) {
-    return reduce(add, iterable) as ReturnValueType<A>;
+    const iterator = iterable[Symbol.asyncIterator]();
+    return iterator.next().then(({ done }) => {
+      if (done) {
+        return 0;
+      }
+
+      return reduce(add, iterable) as ReturnValueType<A>;
+    });
```
hg-pyun, line 43: How to fix this type error...?

image

ppeeou: Thank you for your contribution👍

I left a comment
When we check the return type of `reduce`, we can see that it does not return as we intended.

image

So, if you pass the type as a generic to `reduce`, you can avoid an error.

```ts
   const iterator = iterable[Symbol.asyncIterator]();
     return iterator.next().then(({ done }) => {
       if (done) {
         return 0;
       }

       return reduce<number, number>(add, iterable);
     }) as ReturnValueType<A>;
```

However, if you do this, you can see that type inference does not work well in `sum.test.ts`
I think we need to tell TypeScript a little more about types.

It seems that the singature should be able to receive only the number as shown below.
```ts
function sum<A extends Iterable<number> | AsyncIterable<number>>(
   iterable: A,
): A extends Iterable<number>
   ? number
   : A extends AsyncIterable<number>
   ? Promise<number>
   : never;
```

we decided to take only number, I don't think we need to use `ReturnValueType`.
And the type we will actually implement is as follows.
```ts
function sum<A extends Iterable<number> | AsyncIterable<number>>(
   iterable: A,
): number | Promise<number>
```


Additionally, since we've checked the first value once, we don't need to check the first value again.
```ts
 const iterator = iterable[Symbol.iterator]();
 const { done, value } = iterator.next();
 if (done) {
    return 0;
 }

 return reduce<number, number>((a, b) => add(a, b), value, {
   [Symbol.iterator]() {
      return iterator;
   },
 });
```

shine1594: How about this?
The reduce function is already processing empty values.
```ts
function sum<A extends readonly []>(iterable: A): 0;
function sum<A extends number[]>(arr: A): number;
function sum<A extends Iterable<number>>(iterable: A): number;
function sum<A extends AsyncIterable<number>>(iterable: A): Promise<number>;

function sum<A extends Iterable<number> | AsyncIterable<number>>(
  iterable: A,
): number | Promise<number> {
  if (Array.isArray(iterable)) {
    return iterable.reduce(add, 0);
  } else if (isIterable(iterable)) {
    return reduce<number, number>(add, 0, iterable);
  } else if (isAsyncIterable(iterable)) {
    return reduce<number, number>(add, Promise.resolve(0), iterable);
  }

  throw new TypeError("'iterable' must be type of Iterable or AsyncIterable");
}
```

hg-pyun: Thank you for feedback 🍺

hg-pyun: https://github.com/marpple/FxTS/pull/169/commits/02d23eed89df97ed88a7c66041b7a8205f652bcc

type-check/sum.test.ts
```diff
@@ -5,19 +5,13 @@ const { checks, check } = Test;
...
-const res7 = pipe(["a", "b"], toAsync, sum);
+const res2 = sum(toAsync([1, 2]));
+const res3 = pipe([1, 2], sum);
+const res4 = pipe([1, 2], toAsync, sum);
 
 checks([
   check<typeof res1, number, Test.Pass>(),
-  check<typeof res2, string, Test.Pass>(),
+  check<typeof res2, Promise<number>, Test.Pass>(),
   check<typeof res3, Promise<number>, Test.Pass>(),
```
ppeeou, line 15: `res3` should be `number` type not `Promise<number>`

hg-pyun: oops... if the type is `number`, does not pass type check...

image


Is it needs changed this code?
```ts
function sum<A extends readonly []>(iterable: A): 0;
function sum<A extends number[]>(arr: A): number;
function sum<A extends Iterable<number>>(iterable: A): number;
function sum<A extends AsyncIterable<number>>(iterable: A): Promise<number>;

function sum<A extends Iterable<number> | AsyncIterable<number>>(
  iterable: A,
): number | Promise<number> {
  if (Array.isArray(iterable)) {
    return iterable.reduce(add, 0);
  } else if (isIterable(iterable)) {
    return reduce(add, iterable) as ReturnValueType<A>;
  } else if (isAsyncIterable(iterable)) {
    return reduce<number, number>(add, Promise.resolve(0), iterable);
  }

  throw new TypeError("'iterable' must be type of Iterable or AsyncIterable");
}
```

shine1594: I'll check it out too.

hg-pyun: The syntax does not seem to work properly with type inference.
```ts
function sum<A extends Iterable<number>>(iterable: A): number;
function sum<A extends AsyncIterable<number>>(iterable: A): Promise<number>;
```

hg-pyun: why...

ppeeou: @hg-pyun 
As it is a type affected by pipe, as in the comment [here](https://github.com/marpple/FxTS/pull/169#discussion_r945309277)

I think it should be solved in the following way
```ts
function sum<A extends Iterable<number> | AsyncIterable<number>>(
  iterable: A,
): A extends Iterable<number>
  ? number
  : A extends AsyncIterable<number>
  ? Promise<number>
  : never;
```

If you do as above, you can check that type inference works.
```ts
const res0 = sum([]);
const res1 = sum([1, 2]);
const res2 = sum(toAsync([1, 2]));
const res3 = pipe([], sum);
const res4 = pipe([1, 2], sum);
const res5 = pipe([1, 2], toAsync, sum);

checks([
  check<typeof res0, number, Test.Pass>(),
  check<typeof res1, number, Test.Pass>(),
  check<typeof res2, Promise<number>, Test.Pass>(),
  check<typeof res3, number, Test.Pass>(),
  check<typeof res4, number, Test.Pass>(),
  check<typeof res5, Promise<number>, Test.Pass>(),
]);


```

ppeeou: @hg-pyun  yes,  Since we need to first deliver the type signature provided to the user, the internal implementation of reduce type can proceed later.

shine1594 (approved):
✅

ppeeou (approved):

PR #174, opened 2022-10-17
feat: improve type inference for the 'includes' function

### LINKED ISSUE #173: Improving type inference of 'includes' functions
# Suggestion

## 💻 Use Cases
AS-IS
```ts
const value = '' as string;
const tuple = ['a', 'b', 'c'] as const;

if (includes(value, tuple)) {
  console.log(value); // string
}
```

TO-BE
```ts
const value = '' as string;
const tuple = ['a', 'b', 'c'] as const;

if (includes(value, tuple)) {
  console.log(value); // 'a' | 'b' | 'c'
}
```

This can be useful in the following cases.

```ts
const obj = {
  a: '1',
  b: '2',
  c: '3',
} as const;

if (includes(key, keys(obj)) {
  // `key` is inferred to be 'a' | 'b' | 'c'.
  console.log(obj[key]);
} 
```

shine1594:
Fixes #173

### DIFF
--- a/src/includes.ts
+++ b/src/includes.ts
@@ -15,6 +15,8 @@ import { isAsyncIterable, isIterable } from "./_internal/utils";
  */
 function includes<T extends readonly []>(value: unknown, iterable: T): false;
 
+function includes<T, U extends T>(value: T, iterable: Iterable<U>): value is U;
+
 // prettier-ignore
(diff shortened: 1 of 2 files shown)


ppeeou (approved):
LGTM🤩

PR #179, opened 2022-12-02
fix: comment result of `average()` function

ChanhyukPark-Tech:
Thank you for providing a good library.
There was a difference between the test code and the actual annotation, so I fixed it.

[average docs](https://fxts.dev/docs/average)

image

```tsx
// average.spec.ts

 it.each([
      [[1, 2, 3, 4, 5], 3],
      [[], NaN],
    ])(
      "should return the average of the given elements",
      function (nums, result) {
        expect(average(nums)).toEqual(result);
      },
    );

    it("should be able to be used as a curried function in the pipeline", function () {
      const res = pipe([1, 2, 3, 4, 5], average);
      expect(res).toEqual(3);
    });
```

Thank you😎

### DIFF
--- a/src/average.ts
+++ b/src/average.ts
@@ -10,14 +10,14 @@ import { isAsyncIterable, isIterable } from "./_internal/utils";
  * @example
  * ```ts
  * average([]); // NaN
- * average([1, 2, 3, 4, 5]); // 6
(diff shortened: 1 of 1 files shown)


ppeeou (approved):
Thank you for your contribution :) 👍

PR #183, opened 2022-12-30
Update deps & improved `juxt` function type inference

shine1594:
An error occurred before, but it is not an error after fixing it
```ts
const res = juxt([
  (arg: { a: number, b: number }) => arg.a + arg.b,
  (arg: { b: number, c: string }) => `${arg.b} ${arg.c}`
])({ a: 1, b: 2, c: "3" })
```


ppeeou (approved):
lgtm

PR #185, opened 2022-12-31
Feat/180

### LINKED ISSUE #180: User-Defined Type Guards not working properly in `reject` function
# Suggestion
reproduction: https://codesandbox.io/s/magical-snow-wrrz4m?file=/src/index.ts

## 💻 Use Cases
```ts
import { filter, reject, toArray } from "@fxts/core";

type ObjA = { _type: "A"; a: string; b: string };
type ObjB = { _type: "B"; a: number; b: number };

const list = [
  { _type: "A", a: "1", b: "2" },
  { _type: "B", a: 3, b: 4 },
  { _type: "A", a: "1", b: "2" },
  { _type: "B", a: 3, b: 4 }
];

const res = toArray(reject((a): a is ObjB => a._type === "B", list));

console.log(res); // `res` must be inferred to be of type `ObjA[]`
```

shine1594:
Fixes #180

Improved type inference of `reject` function when receiving custom typeguard function as an argument.
```ts
const res = pipe(
  [1, 2, 3, "4", "5", "6", null, undefined],
  reject(isNil), // (a): a is null | undefined => a == null
  reject(isString), // (a): a is string => typeof a === "string"
  toArray,
);
// As-is
// typeof res == (number, string | null | undefined)[]

// To-be
// typeof res == number[]
```

### DIFF
--- a/.commitlintrc
+++ b/.commitlintrc
@@ -1,3 +1,6 @@
 {
-  extends: ["@commitlint/config-conventional"]
+  extends: ["@commitlint/config-conventional"],
+  rules: {
+    "header-max-length": [1, 'always', 110]
+  }
 }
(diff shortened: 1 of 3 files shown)


ppeeou (approved):
lgtm👍

PR #193, opened 2023-02-17
Feat/190

### LINKED ISSUE #190: Improving type inference when using custom type guards
# Suggestion

## 💻 Use Cases

```ts
const list: { type: "A" | "B", value: number }[] = [{ type: "A", value: number }, { type: "B", value: number }];
const llistA = filter((a): a is { type: "A", value: number } => a.type === "A");
```

shine1594:
Fixes #190 

In this PR, type inference has been improved when dealing with `Discriminated Union` types in `partition` and `reject` functions.

The code below is a typical example of the benefits of this enhancement.
```ts
const list = [
  { type: "A", value: 1 },
  { type: "B", value: 2 }
] as { type: "A" | "B"; value: number }[];

const [listA, listB] = partition((a): a is { type: "A"; value: number } => a.type === "A", list);


listA; // { type: "A", value: number }[]
listB;
/*
 * AS-IS: { type: "A" | "B", value: number }[]
 * TO-BE: { type: "B", value: number }[] 
 */
```

And besides the examples above, there are a few other special cases worth noting.

### When the discriminant property is a boolean type
In general, the type of a discriminant property in Typescript is a subtype (subset) of number or a subtype of string. The number type and string type cannot be used as discriminant properties because the number of elements is infinite. However, since boolean has only two elements, it can be used as a discriminant property. So, although it is not a common case, a library can provide type support such as:

```ts
const list = [
  { type: true, value: 1 },
  { type: false, value: 2 }
] as { type: booelan; value: number }[];

const [listA, listB] = partition((a): a is { type: true; value: number } => a.type === true, list);

listA; // { type: true, value: number }[]
listB;
/*
 * AS-IS: { type: booelan, value: number }[]
 * TO-BE: { type: false, value: number }[] 
 */
```

### When there are multiple discriminant properties
What is the most ideal type inference result for `listB` in the example code below?

```ts
const list = [
  { type: "A", value: 1; value2: true },
  { type: "B", value: 2; value2: false }
] as { type: "A" | "B"; value: 1 | 2; value2: boolean }[];

const [listA, listB] = partition(
  (a): a is { type: "A"; value: 1; value2: true } => a.type === "A" && a.value === 1 && a.value2,
  list
);

listA; // { type: "A"; value: 1; value2: true }[]
listB; // ?[]
```

The answer I think is this:

```ts
Array<
  | { type: "A"; value: 1; value2: false }
  | { type: "A"; value: 2; value2: true }
  | { type: "A"; value: 2; value2: false }
  | { type: "B"; value: 1; value2: true }
  | { type: "B"; value: 1; value2: false }
  | { type: "B"; value: 2; value2: true }
  | { type: "B"; value: 2; value2: false }
>
```

This type includes all possible cases from original type except `{ type: "A"; value: 1; value2: true }`. Although this type inference is not completely impossible, it is very likely to cause performance issues when the number of properties is large, 
Therefore I implemented the function to infer `listB` to its original type (`Array<{ type: "A" | "B"; value: 1 | 2; value2: boolean }>`).


ppeeou (approved):
good 👍

PR #196, opened 2023-03-01
feat: add memoize function

### LINKED ISSUE #195: feature `memoize`
# Suggestion

## ⭐ Suggestion

https://github.com/marpple/FxTS/discussions/194

ppeeou:
Fixes #195

Provides the memoize function with the same interface as [lodash memoize](https://lodash.com/docs/4.17.15#memoize) that users are familiar with.

### DIFF
--- a/src/memoize.ts
+++ b/src/memoize.ts
@@ -1,0 +1,40 @@
+/**
+ * Creates a new function that, stores the results of its calculations in a [`Map`](http://ecma-international.org/ecma-262/7.0/#sec-properties-of-the-map-prototype-object).
+ * When the function is called with same input again, it retrieves the cached result instead of recalculating it.
+ * If resolver is provided, it determines the cache key for storing the result based on the arguments provided to the memoized function.
+ * By default, the first argument provided to the memoized function is used as the map cache key
+ *
+ * @example
+ * ```ts
+ * const add10 = (a: number): number => a + 10;
+ *
+ * const memoized = memoize(add10);
+ * console.log(memoized(5)); // 15
+ * console.log(memoized(10)) // 20
+ * console.log(memoized(5)); // 15 (cached)
+ *
+ * memoized.cache.clear(); // clear cache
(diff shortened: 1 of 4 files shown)


src/memoize.ts
```diff
@@ -0,0 +1,40 @@
...
+ * const memoized = memoize(add10);
+ * console.log(memoized(5)); // 15
+ * console.log(memoized(10)) // 20
+ * console.log(memoized(5)); // 15 (cached)
+ *
+ * memoized.cache.clear(); // clear cache
+ * console.log(memoized(5)); // 15 (no cache)
+ * ```
+ */
+function memoize<
```
shine1594, line 20: Depending on the existence of the resolver function, it seems that we can narrow down the type of the function more precisely by overloading the type of the function.

And if the key of memo is an object, it would be better to use WeakMap.
```ts
function memoize<
  F extends (...args: any[]) => any,
  K extends Parameters<F>[0],
  Return extends F & {
    cache: K extends object ? WeakMap<K, ReturnType<F>> : Map<K, ReturnType<F>>;
  },
>(f: F): Return;
function memoize<
  F extends (...args: any[]) => any,
  Resolver extends (...args: Parameters<F>) => any,
  K extends ReturnType<Resolver>,
  Return extends F & {
    cache: K extends object ? WeakMap<K, ReturnType<F>> : Map<K, ReturnType<F>>;
  },
>(f: F, resolver: Resolver): Return;
```

ppeeou: I'm thinking about whether I should generate a runtime error if neither Map nor WeakMap is assigned to the `cache` value.🧐

shine1594: It would be good to check if the wrong cache is assigned to the runtime and assign a new Map (or WeakMap).
```ts
if (not(memoized.cache instanceof WeakMap || memoized.cache instanceof Map)) {
  memoized.cache = key != null && typeof key === "object" ? new WeakMap() : new Map();
}
```

PR #198, opened 2023-03-14
fix: modify to use WeakMap properly in cache of memoize

### LINKED ISSUE #195: feature `memoize`

shine1594:
Fixes #195 
If the return type of the resolver function is object, then the cache should be assigned to a `WeakMap`. Currently only `Map` is used.

### DIFF
--- a/src/memoize.ts
+++ b/src/memoize.ts
@@ -38,8 +38,12 @@ function memoize<
 >(f: F, resolver?: R) {
   const memoized = (...args: Parameters<typeof f>): ReturnType<F> => {
     const key = typeof resolver === "function" ? resolver(...args) : args[0];
-    const cache = memoized.cache;
-
+    const _self = memoized as any;
+    if (_self.cache === undefined) {
+      _self.cache =
+        key != null && typeof key === "object" ? new WeakMap() : new Map();
+    }
+    const { cache } = _self;
     if (!(cache instanceof WeakMap || cache instanceof Map)) {
       throw new TypeError("`cache` should only use `WeakMap`, `Map`");
     }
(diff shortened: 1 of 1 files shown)


src/memoize.ts
```diff
@@ -38,8 +38,12 @@ function memoize<
...
   const memoized = (...args: Parameters<typeof f>): ReturnType<F> => {
     const key = typeof resolver === "function" ? resolver(...args) : args[0];
-    const cache = memoized.cache;
-
+    const _self = memoized as any;
+    if (_self.cache === undefined) {
+      _self.cache =
+        key != null && typeof key === "object" ? new WeakMap() : new Map();
+    }
+    const { cache } = _self;
```
hg-pyun, line 46: `weakMap` does not support the keys(), values(), entries() methods. Therefore, since you can access the map directly from the outside, it would be safer to have a separate interface to manipulate the cache.

shine1594: So, you're suggesting to restrict the usage of WeakMap as a type for cache and just use Map instead, right?

hg-pyun: No, how about providing an abstracted method?

shine1594: Can you please write it in code and suggest it?

hg-pyun: > Therefore, since you can access the map directly from the outside

I think the sentence above that I wrote before is wrong. I think I'm mistaken. I don't think it's a problem because _self is only used internally. 🙏

hg-pyun (approved):

ppeeou (approved):
👍

ppeeou (approved):
👍

PR #199, opened 2023-03-21
Fix(#197) improve the type inference of the groupBy functions

### LINKED ISSUE #197: Improve the type inference of the `groupBy` and `indexBy` functions
## ⭐ Suggestion
It looks like it could be made better by improving the type inference of the `groupBy` function.

## 💻 Use Cases

```ts
const list = [
  { type: "a", value: 1 },
  { type: "b", value: 2 },
  { type: "c", value: 3 },
  { type: "a", value: 4 } ,
  { type: "b", value: 5 },
  { type: "c", value: 6 },
  { type: "a", value: 7 },
  { type: "b", value: 8 },
  { type: "c", value: 9 },
  { type: "a", value: 10 },
] as { type: "a" | "b" | "c"; value: number; }[]

const res = groupBy(a => a.type, list);

/*
AS-IS
typeof res:
{
    a: {
        type: "a" | "b" | "c";
        value: number;
    }[];
    b: {
        type: "a" | "b" | "c";
        value: number;
    }[];
    c: {
        type: "a" | "b" | "c";
        value: number;
    }[];
}
*/

/*
TO-BE
typeof res:
{
    a: {
        type: "a";
        value: number;
    }[];
    b: {
        type: "b";
        value: number;
    }[];
    c: {
        type: "c";
        value: number;
    }[];
}
*/
```

sunrabbit123:
Fixes #197


src/partition.ts
```diff
@@ -99,12 +102,20 @@ function partition<A extends Iterable<unknown> | AsyncIterable<unknown>, B>(
       }
       return `${Boolean(key)}`;
     }, iterable);
-    return [group["true"] || [], group["false"] || []];
+    const some = [group["true"] || [], group["false"] || []];
```
sunrabbit123, line 105:
```suggestion
```

type-check/groupBy.test.ts
```diff
@@ -30,9 +30,40 @@ const res4 = pipe(
...
+  { type: "c", value: 3 },
+  { type: "a", value: 4 },
+  { type: "b", value: 5 },
+  { type: "c", value: 6 },
+  { type: "a", value: 7 },
+  { type: "b", value: 8 },
+  { type: "c", value: 9 },
+  { type: "a", value: 10 },
+] as { type: "a" | "b" | "c"; value: number }[]);
+
```
shine1594, line 45: I think the test cases below should be added.

```ts
type Res = {
  a: {
    type: "a";
    value: number;
  }[];
  b: {
    type: "b";
    value: number;
  }[];
  c: {
    type: "c";
    value: number;
  }[];
};

pipe(list, groupBy(a => a.type)); // Res
groupBy(a => a.type, toAsync(list)); // Promise<Res>
pipe(list, toAsync, groupBy(a => a.type)); // Promise<Res>
```

shine1594 (changes requested):

src/partition.ts
```diff
@@ -78,7 +78,10 @@ function partition<A extends Iterable<unknown> | AsyncIterable<unknown>, B>(
   f: (a: IterableInfer<A>) => B,
 ): (iterable: A) => ReturnPartitionType<A>;
 
-function partition<A extends Iterable<unknown> | AsyncIterable<unknown>, B>(
+function partition<
```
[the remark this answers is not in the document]
sunrabbit123: But, type about partition is incorrect

I think the type implementation is weird
Changes are indispensable, and I will work on them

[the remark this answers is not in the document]
sunrabbit123:
```ts

function partition<
  A,
  L extends A,
  R extends A = A extends object ? ExcludeObject<A, L> : Exclude<A, L>,
>(f: (a: A) => a is L, iterable: Iterable<A>): [L[], R[]];
```
It's terrible

If this is the case, why is the 'ReturnPartition Type' implemented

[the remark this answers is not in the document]
shine1594: The type signature is for the case below. Please check this test case.
https://github.com/marpple/FxTS/blob/856b178c370e8016d46b15b1ee10353d89c60076/type-check/partition.test.ts#L61

You can easily check with this command: `npm run compile:check`


Could you please explain why this type is not correct?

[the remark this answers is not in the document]
sunrabbit123: This type can pass through the immediate test case. However, due to the lack of consistency in the type structure, errors will be easily spat out when linked to the following types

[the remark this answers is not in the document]
shine1594: It would be good for me to understand if you could write an example of a problem caused by a lack of type consistency in a test code.
However, regardless of this issue, I still think there is no reason to modify 'partition' to improve the type-inferencing
 of 'groupBy' and 'indexBy'.
Please include only code for improvement of 'groupBy' and 'indexBy' in this PR, and create a separate PR for issue of 'partition' function.

src/partition.ts
```diff
@@ -90,7 +93,7 @@ function partition<A extends Iterable<unknown> | AsyncIterable<unknown>, B>(
       return partition(f, iterable as any) as ReturnPartitionType<A>;
     };
   }
-
+  iterable;
```
sunrabbit123, line 96:
```suggestion
```

src/partition.ts
```diff
@@ -99,12 +102,19 @@ function partition<A extends Iterable<unknown> | AsyncIterable<unknown>, B>(
       }
       return `${Boolean(key)}`;
     }, iterable);
-    return [group["true"] || [], group["false"] || []];
+    //    ^?
```
sunrabbit123, line 105:
```suggestion
```

src/partition.ts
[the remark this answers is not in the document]
sunrabbit123: Separating the PR is also good, but there was no change in the return type in the commit that had comments. Only the as syntax was added to the return part.

Although it is a temporary measure due to the change in the return type of groupBy, it is essential. Therefore, changes to partition are inevitable in this PR.

sunrabbit123 (commented):
ㄹ

src/partition.ts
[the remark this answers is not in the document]
sunrabbit123: By the way, most of the functions provided by FxTs have the following return types:

Synchronous or asynchronous
List or iterator (ultimately iterable)
And here are the return types for partition:

[L[], R[]] (essentially, true and false parts)
[A[], A[]] (also divided into true and false)
(iterable: A) => ReturnPartitionType<A> (curried type for 2)
These scattered types result in significant maintenance costs. Is there really a need to have many divisions when they all ultimately return the same value and type?

It would be ideal and require the least maintenance cost to unify them under one type (ReturnPartitionType in this case) and consistently modify only that type.

sunrabbit123:
Additionally, please merge my PR using a squash merge. My commit history is messy...

src/partition.ts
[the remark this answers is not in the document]
sunrabbit123: image
Do you want this?

Don't think that it doesn't look pretty because it's a type alias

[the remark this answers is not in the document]
shine1594: That's exactly what I wanted. Thank you for letting me know a good solution.

type-check/groupBy.test.ts
```diff
@@ -30,9 +30,74 @@ const res4 = pipe(
...
+  groupBy((a) => a),
+); // Res
+
+const res11 = groupBy((a) => a, toAsync(list2)); // Promise<Res>
+const res12 = pipe(
+  list2,
+  toAsync,
+  groupBy((a) => a),
+); // Promise<Res>
+
```
shine1594, line 89: I thought of another edge case, please also consider the following case. Perhaps inferring it in the most original form would be the best option.

```ts
const list1 = [
  { type: "a", value: 1, value2: 'a' },
  { type: "b", value: 2, value2: 'b' },
  { type: "c", value: 3, value2: 'c' },
  { type: "a", value: 4, value2: 'a' },
  { type: "b", value: 5, value2: 'b' },
  { type: "c", value: 6, value2: 'c' },
  { type: "a", value: 7, value2: 'a' },
  { type: "b", value: 8, value2: 'b' },
  { type: "c", value: 9, value2: 'c' },
  { type: "a", value: 10, value2: 'a' },
] as { type: "a" | "b" | "c"; value: number; value2: "a" | "b" | "c" }[];

type Res = { type: "a" | "b" | "c"; value: number; value2: "a" | "b" | "c" };

groupBy((a) => a.type, list1); // { "a": Res[]; "b": Res[]; "c": Res[]; } ?

```

sunrabbit123: In my opinion, this case was written incorrectly. Perhaps the intention was to infer that type and value2 are the same. If that is the case, the signal that type and value2 are the same should have been given when declaring the list.

sunrabbit123:
```ts
const list3 = [
  { type: "a", value: 1, value2: 'a' },
  { type: "b", value: 2, value2: 'b' },
  { type: "c", value: 3, value2: 'c' },
  { type: "a", value: 4, value2: 'a' },
  { type: "b", value: 5, value2: 'b' },
  { type: "c", value: 6, value2: 'c' },
  { type: "a", value: 7, value2: 'a' },
  { type: "b", value: 8, value2: 'b' },
  { type: "c", value: 9, value2: 'c' },
  { type: "a", value: 10, value2: 'a' },
] as ({ type: "a"; value: number; value2: "a" } | { type: "b"; value: number; value2: "b" } | { type: "c"; value: number; value2: "c" })[];

groupBy((a) => a.type, list3);
```

sunrabbit123: What do you think? @shine1594

shine1594: I'm sorry for the late reply. Our team is currently going through a very difficult and challenging time due to workload. We ask for your understanding. Continuing on from the previous comment, if the user asserts the type explicitly as you have shown in the code, there will be no problem at all.

What I'm concerned about is when the types of the `type` and `value2` fields accidentally match in the example code you provided. In such cases, it is impossible to determine at compile time which field will be used as the key for groupBy.

Therefore, I suggested that inferring the original type as I did above is probably the best approach. As a clear example, we cannot assert that there will be no value like`{ type: "a", value: 10, value2: 'c' }` in the `list`.

So, I think the following test case should be added in conclusion:

```ts
type State = "a" | "b" | "c";
type Item = { type: State; value: number; value2: State };
const list: Item[] = [
  { type: "a", value: 1, value2: "a" },
  { type: "b", value: 2, value2: "b" },
  { type: "c", value: 3, value2: "c" },
  { type: "a", value: 4, value2: "a" },
  { type: "b", value: 5, value2: "b" },
  { type: "c", value: 6, value2: "c" },
  { type: "a", value: 7, value2: "a" },
  { type: "b", value: 8, value2: "b" },
  { type: "c", value: 9, value2: "c" },
  { type: "a", value: 10, value2: "a" },
];
const res = groupBy((a) => a.type, list);
checks([check<typeof res, { a: Item[]; b: Item[]; c: Item[] }, Test.Pass>()]);
```

sunrabbit123: Obviously, when we get a key, we don't know what it is.

We'll fix and mention it later.

sunrabbit123: @shine1594 
I try to do something and don't understand it, so I leave it.

I understand the problem you're raising, but the

but it conflicts with what the issue says.

shine1594: Alright, I'll merge this PR for now and try to modify it based on the code you wrote to satisfy [this test case](https://github.com/marpple/FxTS/pull/199#discussion_r1161289493). I'm really grateful for your contributions so far.

shine1594 (approved):

PR #201, opened 2023-04-20
feat: support prop

### LINKED ISSUE #200: Functions from Ramda
# Suggestion

## ⭐ Suggestion

At work, we've been looking for a [ramda](https://ramdajs.com) replacement that is able to handle generic iterables. FxTS would be a great fit for us, but looking at our codebase I found some functions that we use frequently, like [prop](https://ramdajs.com/docs/#prop) and [always](https://ramdajs.com/docs/#always), that currently don't have a direct alternative in FxTS. Those have been proved very useful in many functional pipelines. Would you be open to review a PR adding those to the lib?

## 💻 Use Cases

```ts
const name = "John";
const listOfUsers = [{ name: "John", age: 30 }, /* Many other items... */];
const isUser = includes(name, map(prop("name"), listOfPersons));
```

```ts
const excludeNegatives = false;
pipe(
  [-1, 1, 2, 0 ,-3],
  filter(excludeNegatives ? lt(0) : always(true)),
  toArray,
); // [-1, 1, 2, 0 ,-3]
```

ppeeou:
Fixes #200

support using `prop`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -45,6 +45,7 @@ import pick from "./pick";
 import pickBy from "./pickBy";
 import pipe from "./pipe";
 import pipe1 from "./pipe1";
+import prop from "./prop";
 import reduce from "./reduce";
 import size from "./size";
 import some from "./some";
@@ -106,6 +107,7 @@ export {
   pickBy,
   pipe,
   pipe1,
+  prop,
   reduce,
   size,
   some,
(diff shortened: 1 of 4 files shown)


hg-pyun (approved):
LGTM

ppeeou:
It seems that this function will be merged in that pr https://github.com/marpple/FxTS/pull/202

PR #203, opened 2023-04-21
Fix: length must not be undefined

odoldotol:
Concurrent allows undefined values for its length parameter. It was my mistake, but it caused Node to consume all CPU resources. Moreover, it didn't throw any errors, which wasted time trying to find the cause of the issue. Therefore, I strongly believe that this issue must be fixed.

Fixes #

### DIFF
--- a/src/Lazy/concurrent.ts
+++ b/src/Lazy/concurrent.ts
@@ -92,6 +92,10 @@ function concurrent<A>(
   if (length <= 0) {
     throw new RangeError("'length' must be over 0");
   }
+    
+  if (!length) {
+      throw new TypeError("'length' must be type of number");
+  }
 
   if (!isAsyncIterable(iterable)) {
     throw new TypeError("'iterable' must be type of AsyncIterable");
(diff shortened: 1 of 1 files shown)


src/Lazy/concurrent.ts
```diff
@@ -92,6 +92,10 @@ function concurrent<A>(
   if (length <= 0) {
     throw new RangeError("'length' must be over 0");
   }
+    
+  if (!length) {
+      throw new TypeError("'length' must be type of number");
+  }
```
ppeeou, line 98: How about using `isNumber`?

If we use typescript, we can check the type on props.
but, Is it because of the case in the environment where javascript is used?

shine1594, line 98: L88 - L98
```suggestion
  if (!Number.isFinite(length) || length <= 0) {
    throw new RangeError("'length' must be positive integer");
  }
```

odoldotol: Yes, that's right. As mentioned before, It was entirely my mistake to pass 'undefined' as the parameter.
It could have been easily prevented with TypeScript.

Upon further reflection, I'm not sure if my suggestion was necessary.

Thank you!

odoldotol: I love it

ppeeou (approved):

Issue #206, opened 2023-05-24
How about converting Typeguard function parameter type into generic type.


industriously:
# Suggestion

## ⭐ Suggestion

How about converting `Typeguard function`(is~ functions like isUndefined, isArray) into generic functions?

example

```typescript
/**
 * Returns true if `a` is a undefined.
 *
 * @example
 * ```ts
 * isUndefined(undefined); // true
 * isUndefined(2); // false
 * ```
 */
declare function isUndefined<T>(a: T): a is undefined;
export default isUndefined;
```

## 💻 Use Cases

There is an issue with using the isUndefined function and a custom generic function together, which results in the return type being `any`.

but if every is~~ function will converted into generic function, ts can type-check more properly.

example
// now @fxts/core `isUndefined` function type
<img width="348" alt="스크린샷 2023-05-24 오후 11 55 51" src="https://github.com/marpple/FxTS/assets/116284195/a133cefa-e102-4ae7-956f-46d7f061665a">

// if I convert `isUndefined` signature into generic function.
<img width="329" alt="스크린샷 2023-05-24 오후 11 54 52" src="https://github.com/marpple/FxTS/assets/116284195/df78ba2c-26ee-4c85-82b9-48e1c30089a4">

shine1594:
I appreciate your suggestion. I will consider converting the custom-guard functions into generic functions as you proposed. However, it seems that writing it as follows may achieve the desired result:

```ts
function throwIf<T extends (...param: any) => any>(predicate: T) {
  return function <U extends Parameters<T>[0]>(input: U): U {
    if (predicate(input)) {
      throw new Error();
    } else {
      return input;
    }
  };
}
```

industriously:
Thank you for the positive feedback. Adding to my previous response, for the isUndefined function, the following function signature would be appropriate

```typescript
declare function isUndefined<T>(input:T|undefined): input is undefined;
```

ppeeou:
@industriously  I think it would be ok to provide a signature as you wrote
Could you please create a PR??

industriously:
> @industriously  I think it would be ok to provide a signature as you wrote
> 
> Could you please create a PR??

sure!

industriously:
> @industriously I think it would be ok to provide a signature as you wrote Could you please create a PR??

Can I add more function in PR??
- isDefined
- throwIf

PR #207, opened 2023-05-27
refactor: remove any type in isUndefined.

industriously:
- fix(#206): remove any type
- feat: add isDefined function
- feat: add throwIf function


test/throwIf.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { identity, isString } from "../src";
+import throwIf from "../src/throwIf";
```
ppeeou, line 2:
```suggestion
import { identity, isString, throwIf } from "../src";
```

test/isDefined.spec.ts
```diff
@@ -0,0 +1,16 @@
+import isDefined from "../src/isDefined";
```
ppeeou, line 1:
```suggestion
import {isDefined} from "../src";
```

ppeeou: To make sure we are doing well exporting to src
we are writing the code in the form of `../src`

ppeeou (commented):
Thank you for your contribution 

I have a few questions and requests.

- How do you usually use the `throwIf` function?
- Could you please add a type test as well?
  - https://github.com/marpple/FxTS/tree/main/type-check

industriously:
> Thank you for your contribution
> 
> I have a few questions and requests.
> 
> * How do you usually use the `throwIf` function?
> * Could you please add a type test as well?
>   
>   * https://github.com/marpple/FxTS/tree/main/type-check

I am a backend developer. In backend environments, when possible, I write function logic using `pipe`. In the following cases, I use throwIf:

- If data is retrieved from the database but it is not found, I throw a NotFound error.
- If the return type of an internal utility function represents a failure case, I throw an appropriate error for it.
I take inspiration from Rust's Result<T, E> enum and design my code so that internal functions return failure cases instead of directly throwing errors.


```typescript
export namespace Service {
  export const getOne = <T, R>(input: {
    user_id: string;
    findFirst: (id: string) => Promise<T | null>;
    exception_for_notfound: unknown;
    validator: (input: T) => T;
    mapper: (arg: T) => R;
  }) =>
    pipe(
      input.user_id,

      input.findFirst,

      throwIfNull(() => input.exception_for_notfound), // throwIfNull is throwIf(isNull)

      input.validator,

      input.mapper
    );
}

```

src/isEmpty.ts
```diff
@@ -24,39 +27,21 @@
...
-    return false;
-  }
-
-  if (typeof value === "undefined" || value === null) {
-    return true;
-  }
-
-  if (value instanceof Date) {
-    return false;
-  }
```
ppeeou, line 38: It looks like this logic is missing

I think it would be better not to change the isEmpty logic. Is there a reason to change it?

industriously: I used to think that an Empty Object refers to a literal object that does not have any properties at all. Objects like Date or other custom class instances can have private attributes. These attributes may not be accessible, and if there are no public properties, they can deceive others into thinking that the object is empty. However, class instances like Date can internally hold data and have some meaningful content even without public properties.

Therefore, I changed the logic to consider an object as an Empty Object and return true only if it is a literal object with Object as its constructor, meaning it is not a class instance and has no methods or properties.

Is my understanding of the Empty Object definition different from what you had in mind?

Note: The code I wrote, including Date, returns false for all class instances.

----
in korea,

제가 이해한 Empty Object의 정의는 클래스 인스턴스가 아닌 리터럴 객체이고 메소드나 속성이 전혀없는 빈 객체 "{}" 를 생각했습니다. 그 예로 Date 클래스 인스턴스는 public 속성이 전혀 없어서 프로퍼티가 전혀없는 빈 객체인 것처럼 보이지만 내부 데이터로 특정 시간 정보를 지니고 있습니다.

그래서 제가 작성한 코드는 정확히 {} 객체만을 구별해서 true를 반환하도록 수정했습니다. 혹시 제가 `@fxts/core`에서 정의한 Empty Object에 대해 오해하고 있나요?
일단 현재 제가 작성한 코드는 모든 테스트를 통과합니다.

ppeeou: I think you are right 👍

Since we use "squash and merge", it would be better to add the corresponding commit separately

industriously: ok, i will create new pr for isEmpty!

ppeeou (approved):
lgtm

src/throwIf.ts
```diff
@@ -0,0 +1,29 @@
...
+ *    throwIf(isUndefined)((input) => Error("return of fn is undefined")),
+ *    // input is undefined, and it is return of fn
+ *
+ *    (input) => input, // input is string
+ *  )
+ * ```
+ *
+ * @param predicate typeguard function for T
+ * @return T exclude N
+ */
```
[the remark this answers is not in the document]
industriously: Considering usability, it seems like predicates should return a boolean. This is because JavaScript allows implicit boolean conversions for all values in if statements, but the results are inconsistent. Due to this, and the fact that understanding JavaScript's language-specific features is required to predict the outcome accurately, it would be better to restrict predicates to functions that return boolean (or use type guards).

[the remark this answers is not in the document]
industriously: Also, considering that the argument 'err' accepts "input" as its parameter, would it be beneficial to use the `identity` function as the default value?

Using the identity function as the default value could be a good option.

src/isDefined.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Returns true if `a` is not a undefined.
+ *
+ * @example
+ * ```ts
+ * isDefined(undefined); // false
+ * isDefined(null) // true
+ * isDefined(2); // true
+ * ```
+ */
+export const isDefined = <T>(a: T | undefined): a is T => !isUndefined(a);
```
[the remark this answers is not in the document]
industriously: That's a great idea!

src/throwIf.ts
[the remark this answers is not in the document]
shine1594: Alright. I agree with your opinion, and anyway, the main point of my suggestion is to allow variadic arguments. It seems better to continue working on this in a separate PR, apart from this one. In this PR, please only keep the code related to the support of generic types in custom-typeguard functions, and please remove the code for adding new functions (`isDefined` and `throwIf`) for now.

ppeeou (approved):

shine1594 (approved):

PR #208, opened 2023-05-30
refactor: change isEmpty logic

industriously:
if value is object, it have to literal object and have no method and property.
if value is instance of class, isEmpty always return true. ex) Date, Custom Class, .....
- but array instance is considered to have item or not.


ppeeou (approved):
lgtm

shine1594 (approved):

PR #219, opened 2023-07-10
feat: unless, throwError

industriously:
add unless function
add throwError function

this pr is about #215


src/types/ReturnPipeType.ts
```diff
@@ -1,27 +1,35 @@
...
+  : T["length"] extends 1
+  ? Head<T>
+  : Awaited<T[1]> extends never
+  ? never
+  : PipeLast<Tail<T>>;
+
 type ReturnPipeType<
   T extends any[],
-  R = Awaited<Last<T>>,
+  R = Awaited<PipeLast<T>>,
```
shine1594, line 32: Could you please write a test case in the `pipe.test.ts` file to verify what improvements this change has made?

industriously: 로직 중간에 never 타입이 존재하면, 이후 함수를 고려하지 않고 최종 타입이 never(만약 비동기 작업을 포함한다면 Promise<never>)가 연산되도록 수정했습니다.
그에 대한 케이스도 추가했습니다!

ppeeou (approved):
lgtm

shine1594 (commented):
Personally, I don't think it's easy to infer what a function does just by looking at the name `skip`. In Ramda, there is a function called `unless` that performs the same action, and I think this name is more intuitive. What are your thoughts on changing the function name to [`unless`](https://ramdajs.com/docs/#unless)? @industriously @ppeeou

industriously:
> Personally, I don't think it's easy to infer what a function does just by looking at the name `skip`. In ramda.js, there is a function called `unless` that performs the same action, and I think this name is more intuitive. What are your thoughts on changing the function name to [`unless`](https://ramdajs.com/docs/#unless)? @industriously @ppeeou

i agree!

shine1594 (approved):
👍

ppeeou:
> > Personally, I don't think it's easy to infer what a function does just by looking at the name `skip`. In ramda.js, there is a function called `unless` that performs the same action, and I think this name is more intuitive. What are your thoughts on changing the function name to [`unless`](https://ramdajs.com/docs/#unless)? @industriously @ppeeou
> 
> i agree!


I agree too
Could you fix the `unless` document as well? @industriously

industriously:
> > > Personally, I don't think it's easy to infer what a function does just by looking at the name `skip`. In ramda.js, there is a function called `unless` that performs the same action, and I think this name is more intuitive. What are your thoughts on changing the function name to [`unless`](https://ramdajs.com/docs/#unless)? @industriously @ppeeou
> > 
> > 
> > i agree!
> 
> I agree too Could you fix the `unless` document as well? @industriously

i commited. please check it.

ppeeou:

Issue #218, opened 2023-06-29
Pipe as lazy function


darky:
# Suggestion

## ⭐ Suggestion

`pipe` in **FxTS** works in eager fashion.

```ts
pipe(1, (n) => n + 1) // receiving of 2 immediately
```

Would be great to have something like `pipeLazy`, which works in lazy fashion.

```ts
const fn = pipeLazy((n: number) => n, n => n + 1) // Function<(n: number) => number>
fn(1) // 2
```

## 💻 Use Cases

* Huge amount of FP libraries, that I know, using lazy pipe, not eager
  * [lodash.flow](https://lodash.com/docs/4.17.15#flow)
  * [ramda.pipe](https://ramdajs.com/docs/#pipe)
  * [rxjs.pipe](https://rxjs.dev/api/index/function/pipe)
  * [rocket.pipes](https://github.com/darky/rocket-pipes)
  ...
* Lazy pipes more JIT friendly and faster. They can be initialized once on application bootstrap and will works with maximum speed. Eager pipes will be generated every time and spend additional CPU resources.
* Lazy pipes can be easily composed with each other.
* Lazy pipes reduces pipe hell. Lazy pipes propagates to creating many little pipes. Eager pipes propagates to creating one big pipe.

ppeeou:
Good idea. I already checked your library rocket.pipes 👍

Could you please create a PR??

darky:
> Good idea. I already checked your library rocket.pipes 👍
> 
> Could you please create a PR??

Thanks, yep 👌 
But need to wait some time, when I will get free time 🕐

ppeeou:
@darky  There is something I would like to discuss.

Currently, `pipeLazy` supports code type 1, but code type 2 is not supported. Existing functions used in `FxTS` are designed to receive and determine whether it is Iterable or AsyncIterable. In the case of shape 2, it is not included in the specified signature because we do not know what type will come from the `head` function.

1. pass
```ts
const res1 = pipeLazy(
    (a: string) => head(a),
);
```
2. fail
```ts
const res2 = pipeLazy(
   head,
);
```


So the initial design was intended for users to type something like the following to use the `pipeLazy` function.
```ts
const res3 = (a:string) => pipe(a, head)
```

Is there any way to support type 2?🧐

darky:
@ppeeou 

Sorry, but I don't think that case 2 is possible. 
[...]

PR #220, opened 2023-07-19
feat: pipeLazy (#218)

darky:


src/Lazy/pipeLazy.ts
```diff
@@ -0,0 +1,333 @@
...
+ * // with toAsync
+ * await pipeLazy(
+ *  toAsync,
+ *  map(a => a + 10),
+ *  filter(a => a % 2 === 0),
+ *  toArray,
+ * )([Promise.resolve(1), Promise.resolve(2), Promise.resolve(3), Promise.resolve(4), Promise.resolve(5)]); // [12, 14]
+ * ```
+ *
+ * {@link https://codesandbox.io/s/fxts-toarray-fy84i | Try It}
```
darky, line 41: @ppeeou @shine1594 can you please fix/correct only this part (codesandbox, doc, doc link)

ppeeou: sure :)

Could you add type test? 
you can see reference https://github.com/marpple/FxTS/tree/main/type-check

darky: > Could you add type test? you can see reference https://github.com/marpple/FxTS/tree/main/type-check

✅

darky: @ppeeou if you have another notices, I can fix it too

ppeeou:
@darky Thank you for your contribution
sorry for the late confirmation

I left a comment
In the following situation, function inference is wrong
Is it possible to implement a function to be inferred as a 'Promise' ?

```ts
const res = pipeLazy((a: number) => String(a))(Promise.resolve(2));
// infer res -> String not Promise(String)
```

darky:
> I left a comment In the following situation, function inference is wrong Is it possible to implement a function to be inferred as a 'Promise' ?
> 
> ```ts
> const res = pipeLazy((a: number) => String(a))(Promise.resolve(2));
> // infer res -> String not Promise(String)
> ```

@ppeeou fixed

ppeeou (approved):
LGTM👍

PR #229, opened 2023-09-25
fix(sort): replace with internal sort

hg-pyun:
The `SortBy` is completely same `Sort`. Therefore, remove duplicate and using internal implementation.


ppeeou (approved):
lgtm

PR #236, opened 2023-10-09
feat: change `reduce` behavior, when empty iterable

### LINKED ISSUE #235: Change `reduce` behavior, When there are two arguments for `reduce`, and the first iterator value is done:true
### suggestion
If there are three arguments for `reduce`, there is no issue.
In the case of two arguments for `reduce` and the first iterator value is done:true, I think it should throw an error

### why?
When there are two arguments for reduce, and the first iterator value is done:true, it creates an ambiguous situation regarding the return value. 
If we choose to pass `undefined` in this case, the return type should also reflect the potential presence of `undefined` type ex) `T | undefined`

### other case
This behavior also applies to `Array.prototype.reduce`, where the same error will occur.
image

Also, for Iterator Helpers `reduce`, in the mentioned scenario, a TypeError should be raised   in https://tc39.es/proposal-iterator-helpers/#sec-iteratorprototype.reduce.

ppeeou:
Fixes #235

### DIFF
--- a/src/every.ts
+++ b/src/every.ts
@@ -58,7 +58,12 @@ function every<
     return pipe(
       map(f, iterable),
       takeUntil(not),
-      reduce((a, b) => a && b),
+      (acc) =>
+        reduce(
+          (a: boolean, b: boolean) => a && b,
+          true,
+          acc as Iterable<boolean>,
+        ),
       (a) => a ?? true,
       Boolean,
     );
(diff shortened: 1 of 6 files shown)


hg-pyun (approved):
If the code is previously operating as undefined, it can be a breaking change. I think it would be good to mention this in the next release note.

ppeeou:
> If the code is previously operating as undefined, it can be a breaking change. I think it would be good to mention this in the next release note.

yes I will

shine1594 (approved):
LGTM 👍

PR #243, opened 2024-01-21
fix(reduce): better type and docs for `reduce` (#240)

### LINKED ISSUE #240: Better documentation and parameter name for `reduce`
# Suggestion

### Bad parameter name or generic name

In the jsdoc of `reduce`, variable name is really confusing since it's just `a` and `b`. After I inspect type signature, it's getting worse because of `(a: B, b: A) => B`. 

In tslib, they used following for `Array.prototype.reduce` .

```ts
reduce<U>(callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: T[]) => U, initialValue: U): U;
```

In [#MDN reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce), they use following description.

> Its return value becomes the value of the `accumulator` parameter on the next invocation of `callbackFn`.

For my suggestion, how about this? (in the manner of FxTS convention which has mathematical short things)

```ts
reduce<T, R>(f: (acc: R, value: T) => R, seed: R, ...)
```

### Requires more elaboration for docs of `reduce`

I think It's good to explain case of accumulator differ to value. I think it's pretty common in real life. Like adding following in the jsdoc of `reduce`.

```ts
// with accumulator
const data = [{ name: 'ppeeou', score: 25751 }, { name: 'shine1594': score: 74317 }, { name: 'jbl428', score: 366 }]
const scoreSum = reduce((acc, { score }) => acc + score, 0, data)
```

## 💻 Use Cases

More developers can understand the powerful features of `reduce`.

Phryxia:
At first I mentioned`R` but more describing `Acc` would be better.

- Change type `A`,`B` into `T`, `Acc` for `reduce`
- Give proper value name for `acc`, `value` instead of `a`, `b`
- Add example with explicit seed
- Add `@typeParam` for [tsdoc](https://tsdoc.org/pages/tags/typeparam/) support

Fixes #240

### DIFF
--- a/src/reduce.ts
+++ b/src/reduce.ts
@@ -4,21 +4,25 @@ import type Arrow from "./types/Arrow";
 import type IterableInfer from "./types/IterableInfer";
 import type ReturnValueType from "./types/ReturnValueType";
 
-function sync<A, B>(f: (a: B, b: A) => B, acc: B, iterable: Iterable<A>): B {
+function sync<T, Acc>(
+  f: (a: Acc, b: T) => Acc,
+  acc: Acc,
+  iterable: Iterable<T>,
+): Acc {
   for (const a of iterable) {
     acc = f(acc, a);
   }
   return acc;
 }
 
-async function async<A, B>(
-  f: (a: B, b: A) => B,
-  acc: Promise<B>,
-  iterable: AsyncIterable<A>,
+async function async<T, Acc>(
+  f: (a: Acc, b: T) => Acc,
+  acc: Promise<Acc>,
+  iterable: AsyncIterable<T>,
 ) {
   for await (const a of iterable) {
     // becauseof using es5, use `await`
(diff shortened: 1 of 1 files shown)


src/reduce.ts
```diff
@@ -27,19 +31,49 @@ async function async<A, B>(
...
  *  reduce(sum),
  * ); // 26
+ * ```
+ *
+ * Currently, type with explicit seed form can't be inferred properly due to the limitation of typescript.
+ * But you can still use mostly with @ts-ignore flag. For more information please visit issue.
  *
+ * {@link https://github.com/marpple/FxTS/issues/239 | #related issue}
+ *
+ * Asynchronous things don't matter.
```
ppeeou, line 74:
```suggestion
```

ppeeou (approved):

PR #245, opened 2024-02-17
feat: implement `reduceLazy` (#239)

### LINKED ISSUE #239: Missing type of `reduce` with initial seed in declarative form
# Bug Report

Consider following scenario.

```
sample x, y from R^3 (ex: [1, 2, 3], [4, 5, 6])
compute inner product of <x, y>
```

Currently, with FxTS, it can be done with following pipeline

```ts
function dot(a: Iterable<number>, b: Iterable<number>) {
  return pipe(
    zip(a, b),
    map(([a, b]) => a * b),
    reduce((acc, x) => acc + x),
  )
}
```

But actually, I tried with following approach at first, which has type failure. ~~Interestingly this works if I bypass type check. And it seems to be natural to do so, similar to other FxTS APIs.~~ I think I was wrong at that time for some unknown reasons. It is not supported in current version. (added on 2024-02-17)

```ts
function dot2(a: Iterable<number>, b: Iterable<number>) {
  return pipe(
    zip(a, b),
    reduce((acc, [x, y]) => acc + x * y, 0), // <----- no appliable signature
  )
}
```

### 🙁 Actual behavior

There is no such type signature for declarative form of `reduce` with `seed` in the type definition.

### 🙂 Expected behavior

Maybe, this might work. If you consider this is valid, tell me so I'll raise for pull request.

```ts
function reduce<
  A extends Iterable<unknown> | AsyncIterable<unknown>, 
  B
>(f: (a: B, b: IterableInfer<A>) => B | Promise<B>, seed: B): (iterable: A) => ReturnValueType<A, B>;
```

---

### Version Information

- Operating system: Windows 10 
- Typescript: TypeScript 4.8
- Nodejs: v16.16.0

Phryxia:
Fixes #239 

- Extract type of function argument in `reduce` as `SyncReducer` and `AsyncReducer`
   - This is because they are both used in `reduce` and `reduceLazy`
   - It's more concise than messy raw arrowy type
- Implement `reduceLazy`
   - At first I thought the implementation of `reduce` covers lazy cases 100%, but it wasn't
   - To resolve issue just relay `reduce` with rich dedicated types
- Add type test and spec test for `reduceLazy`
   - Type tests were desigend using exhaustive cartessian thinking of following three dimension:
      - *homogeneous* vs *heterogeneous*
      - `SyncReducer` vs `AsyncReducer`
      - `Iterable` vs `AsyncIterable`
   - Spec tests were almost identical to `reduce`
      - Test for promise rejection was omitted since it is responsible for `reduce`

### DIFF
--- a/src/Lazy/reduceLazy.ts
+++ b/src/Lazy/reduceLazy.ts
@@ -12,6 +12,50 @@ type InferCarrier<T> = T extends AsyncIterable<infer R>
 // DO NOT change the order of signatures prematurely.
 // See `reduceLazy.test.ts` for the reason
 
+/**
+ * High order functional version of `reduce`, which behaves identical to it.
+ *
+ * @param f Reducer function `(acc, value) => acc`. It can be both synchronous and asynchronous.
+ * @param seed Initial value. Note that if the type of `acc` and `value` differ, `seed` must be given.
+ *
+ * @example
+ * Type must be provided for stand alone call.
+ *
+ * ```ts
+ * const reduce = reduceLazy((a: number, b: number) => a + b, 5)
+ *
+ * reduce([1, 2, 3]) // number
+ * reduce(toAsync([1, 2, 3])) // Promise<number>
+ * ```
+ *
+ * Fit perfectly with `pipe`
+ *
+ * ```ts
+ * pipe(
+ *   [1, 2, 3, 4],
+ *   reduceLazy((a, b) => a + b, 5)
+ * ); // 15
+ * ```
+ *
+ * You can use asynchronous callback
+ *
+ * ```ts
+ * await pipe(
+ *   [1, 2, 3, 4],
+ *   reduceLazy(async (a, b) => a + b, 5)
+ * ); // 15
+ * ```
+ *
+ * `AsyncIterable` doesn't matter.
+ *
+ * ```ts
+ * await pipe(
+ *   [1, 2, 3, 4],
+ *   toAsync,
+ *   reduceLazy((a, b) => a + b, 5)
+ * ); // 15
+ * ```
+ */
 function reduceLazy<T extends Iterable<unknown> | AsyncIterable<unknown>, Acc>(
   f: SyncReducer<Acc, IterableInfer<T>> | AsyncReducer<Acc, IterableInfer<T>>,
   seed: Acc,
--- a/src/Lazy/reduceLazy.ts
+++ b/src/Lazy/reduceLazy.ts
@@ -1,0 +1,63 @@
+import reduce from "../reduce";
+import type IterableInfer from "../types/IterableInfer";
+import type { AsyncReducer, SyncReducer } from "../types/Reducer";
+import type ReturnValueType from "../types/ReturnValueType";
+
+type InferCarrier<T> = T extends AsyncIterable<infer R>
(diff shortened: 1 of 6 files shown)


src/Lazy/reduceLazy.ts
```diff
@@ -0,0 +1,63 @@
...
+type InferCarrier<T> = T extends AsyncIterable<infer R>
+  ? AsyncIterable<R>
+  : T extends Iterable<unknown>
+  ? T
+  : never;
+
+// DO NOT change the order of signatures prematurely.
+// See `reduceLazy.test.ts` for the reason
+
+function reduceLazy<T extends Iterable<unknown> | AsyncIterable<unknown>, Acc>(
```
Phryxia, line 15: I think I forgot jsdoc for this. I'll do it right away.

src/Lazy/reduceLazy.ts
```diff
@@ -0,0 +1,63 @@
...
+  f: SyncReducer<Acc, IterableInfer<T>>,
+  seed?: Acc,
+) {
+  if (seed === undefined) {
+    return (iterable: Iterable<unknown> | AsyncIterable<unknown>) =>
+      reduce(f, iterable as any);
+  }
+  return (iterable: Iterable<unknown> | AsyncIterable<unknown>) =>
+    reduce(f, seed, iterable as any);
```
Phryxia, line 60: I don't have any idea to overcome this, and `reduce` also uses `as any` so I just simply suppress them.
Actually it's common to use anyscript(😆) for implementation.

src/reduce.ts
```diff
@@ -114,48 +115,42 @@ function reduce<T extends readonly [], Acc>(
...
 ): Promise<Acc>;
 
 function reduce<T extends Iterable<unknown> | AsyncIterable<unknown>>(
-  f: (
-    acc: IterableInfer<T>,
-    value: IterableInfer<T>,
-  ) => IterableInfer<T> | Promise<IterableInfer<T>>,
-): (iterable: T) => ReturnValueType<T, IterableInfer<T>>;
+  f: AsyncReducer<IterableInfer<T>, IterableInfer<T>>,
+): (iterable: T) => ReturnValueType<T>;
```
Phryxia, line 146: `ReturnValueType<T>` is equivalent to `ReturnValueType<T, IterableInfer<T>>`. See `ReturnValueType.ts`.

Phryxia:
Note that test succeed for `reduce` and `reduceLazy`. CI failed because of #93

test/Lazy/reduceLazy.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// these tests are almost identical to `reduce.spec.ts`
+describe("reduceLazy", () => {
+  describe("sync", () => {
+    it("should return initial value when the given `iterable` is empty array", () => {
+      const reduce = reduceLazy((a, b) => a + b, "seed");
+      expect(reduce([])).toEqual("seed");
+    });
+
+    it("should be occured error when the given `iterable` is an empty array and initial value is absent", () => {
```
ppeeou, line 14:
```suggestion
    it("should be occurred error when the given `iterable` is an empty array and initial value is absent", () => {
```

ppeeou:
LGTM👍

PR #253, opened 2024-03-09
feat: key support bigint

ppeeou:
Fixes #

Bigint can be assigned as a key to an object, so add it to the Key type.


ppeeou:
Index signature parameter type does not support bigint😭

PR #271, opened 2024-04-25
feat: add github source code link in api docs

hg-pyun:
# Summary
Add a code to link the source code so that you can see it directly in the API document.

## Image
image


website/markdown-plugin/src/MarkdownFeature.ts
```diff
@@ -7,6 +7,8 @@ import {
 } from "@microsoft/api-documenter";
 import { ApiItem } from "@microsoft/api-extractor-model";
 
+import FunctionJson from "./function.json"; // Same file as /website/function.json. Used to judge path inside markdown-plugin, copy inside because of file path issue after build.
```
hg-pyun, line 10: It's possible that this part will become a debt in the long run. When new methods are added, we need to fix both places, so we'll make sure that the improvement in this part is separately PR.

ppeeou (approved):
👍👍

PR #272, opened 2024-04-25
fix: dropUntil.ts example

hg-pyun:
## As-Is
```ts
// The example answer is 3, 4, 5, but the correct answer is 2, 3, 4, 5, 1, 2 
const iter = dropUntil((a) => a < 3, [1, 2, 3, 4, 5, 1, 2])
```

## To-Be
I think it is a mistake copying and pasting `dropWhile`. Considering the context, I think it's the right way to change it as follows.
```ts
const iter = dropUntil((a) => a > 3, [1, 2, 3, 4, 5, 1, 2]);
```

### DIFF
--- a/src/Lazy/dropUntil.ts
+++ b/src/Lazy/dropUntil.ts
@@ -70,10 +70,10 @@ function async<A, B>(
  *
  * @example
  * ```ts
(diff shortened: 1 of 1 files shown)


ppeeou (approved):
lgtm

PR #282, opened 2024-09-11
docs: example sort

### LINKED ISSUE #281: sort example not working
# Bug Report

- Sort example is not working
- https://fxts.dev/docs/sort
 
```
sort((a, b) => a > b, [3, 4, 1, 2, 5, 2]); // [1, 2, 2, 3, 4, 5]
sort((a, b) => a > b, 'bcdaef); // ["a", "b", "c", "d", "e", "f"]
```

- but when i changed sort function like this work well
```
sort((a, b) => a - b, [3, 4, 1, 2, 5, 2]); // [1, 2, 2, 3, 4, 5]
```

### 💻 Code

- sort use Javscript Array sort method
https://github.com/marpple/FxTS/blob/main/src/sort.ts#L57C11-L57C15 
- Javascript Array sort method does not work like the docs example
스크린샷 2024-09-10 오후 6 57 04

```ts

```

### 🙁 Actual behavior

```
sort((a, b) => a > b, [3, 4, 1, 2, 5, 2]); // [3, 4, 1, 2, 5, 2]
sort((a, b) => a > b, 'bcdaef); // ['b', 'c', 'd', 'a', 'e', 'f']
```

### 🙂 Expected behavior

```
sort((a, b) => a > b, [3, 4, 1, 2, 5, 2]); // [1, 2, 2, 3, 4, 5]
sort((a, b) => a > b, 'bcdaef); // ["a", "b", "c", "d", "e", "f"]
```

I don't know what the intent is

If it is intended to use '>' inside the Sort function, then the function should be fixed.
If not, I think you need to fix the example

---

### Version Information

- Operating system:
- Typescript:
- Nodejs:

ppeeou:
Fixes #281

### DIFF
--- a/src/sort.ts
+++ b/src/sort.ts
@@ -11,7 +11,7 @@ import type ReturnValueType from "./types/ReturnValueType";
  * @example
  * ```ts
  * sort((a, b) => a > b, [3, 4, 1, 2, 5, 2]); // [1, 2, 2, 3, 4, 5]
- * sort((a, b) => a > b, 'bcdaef); // ["a", "b", "c", "d", "e", "f"]
+ * sort((a, b) => a.localeCompare(b), "bcdaef"); // ["a", "b", "c", "d", "e", "f"]
  * ```
  */
 // prettier-ignore


src/sort.ts
```diff
@@ -11,7 +11,7 @@ import type ReturnValueType from "./types/ReturnValueType";
  * @example
  * ```ts
  * sort((a, b) => a > b, [3, 4, 1, 2, 5, 2]); // [1, 2, 2, 3, 4, 5]
```
flex-hyuntae, line 13: This example also not working

PR #292, opened 2024-10-25
fix: fix incorrect active color in navigation bar

guesung:
Fix incorrect active color in navigation bar

Previously, the navigation items in both doc and API sections were highlighted simultaneously.
This PR separates the active states by.
- Using `docId` to determine which item should be highlighted

This ensures only the correct navigation item is highlighted at a time.

# Before

스크린샷 2024-10-25 오후 1 51 53

# After

스크린샷 2024-10-25 오후 1 52 06

## In video

https://github.com/user-attachments/assets/25c682c3-0fec-4bdd-a8de-57b5fb1d0072

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -57,8 +57,8 @@ const config = {
           src: "img/fxts.png",
         },
         items: [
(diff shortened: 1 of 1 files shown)


ppeeou:

PR #306, opened 2025-02-27
feat: add `concurrentPool`

### LINKED ISSUE #304: Q. Is there a good way to implement concurrency using a queue-like approach instead of the `concurrent` function?
Q. When implementing concurrency for asynchronous functions, the documentation recommends using `concurrent`.

But, the problem with both `concurrent` and `Promise.all` is that even if some functions finish quickly, the overall execution time is still determined by the slowest asynchronous function.

I’d like to implement, once some asynchronous functions finish early, the remaining pending functions are immediately queued—ensuring that the desired concurrency level is consistently maintained throughout the entire loop over the collection.

To achieve this, I am currently using an external library like [p-queue](https://github.com/sindresorhus/p-queue), as shown in the example below.

```ts
import PQueue from "p-queue"

const concurrency = 3
const q = new PQueue({ concurrency })
const now = performance.now()

pipe(
  range(Number.MAX_SAFE_INTEGER),
  take(99),
  map((i) => async () => {
    await delay(i * 100)
    console.log(i, Math.floor(performance.now() - now) + "ms")
    return i
  }),
  map(q.add.bind(q)),
  toArray,
)

// 0 3ms
// 1 102ms
// 2 202ms
// 3 305ms
// 4 503ms
// 5 703ms
// 6 905ms
// 7 1204ms
// 8 1504ms
// 9 1806ms
// 10 2205ms
// 11 2605ms
// 12 3007ms
// ...
```

compare to:

```ts
await pipe(
  range(Number.MAX_SAFE_INTEGER),
  toAsync,
  map(async (i) => {
    await delay(i * 100)
    console.log(i, Math.floor(performance.now() - now) + "ms")
    return i
  }),
  take(999),
  concurrent(3),
  toArray,
)

// 0 1ms
// 1 101ms
// 2 201ms
// 3 503ms
// 4 604ms
// 5 703ms
// 6 1305ms
// 7 1405ms
// 8 1505ms
// 9 2406ms
// 10 2506ms
// 11 2607ms
// 12 3809ms
// ...
```

Is there a native way in `@fxts/core` to accomplish this without relying on external libraries like `p-queue`?

---

Q. The `range` function accepts `Infinity` as an argument, but why can’t the `concurrent` function accept `Infinity` (like same behavior with `Promise.all`)?
Is this an intentional design decision, or is there room for improvement?

ppeeou:
Fixes #304

I'd like to get feedback from you  @ysm-dev 
What do you think about the expression and behavior?

### DIFF
--- a/src/Lazy/concurrentPool.ts
+++ b/src/Lazy/concurrentPool.ts
@@ -11,7 +11,17 @@ type Item<T> = {
 };
 
 /**
- * @experimental concurrentPool
+ * Ensures consistency in the level of concurrency based on the given number of concurrent tasks.
+ *
+ * @example
+ * ```ts
+ * await pipe(
+ *   [1, 2, 3, 4, 5, 6],
+ *   toAsync,
+ *   map((a) => delay(100 * a, a)),
+ *   concurrentPool(3),
+ *   each(console.log), // log 1, 2, 3, 4, 5, 6
+ * ); // 2 seconds
  */
 function concurrentPool<A>(
   length: number,
--- a/src/Lazy/concurrentPool.ts
+++ b/src/Lazy/concurrentPool.ts
@@ -46,9 +46,6 @@ function concurrentPool<A>(
     return map(identity, iterable);
   }
 
-  /**
-   * @TODO ppeeou: Considering id circulation
-   */
   let fillId = 1;
   let consumeId = 1;
   let idleWorkers = length;
@@ -67,6 +64,7 @@ function concurrentPool<A>(
 
     idleWorkers--;
     const id = fillId++;
+    fillId = fillId % Number.MAX_SAFE_INTEGER;
     try {
       const item = {
         success: await iterator.next(),
@@ -161,6 +159,7 @@ function concurrentPool<A>(
     async next() {
       return new Promise((resolve, reject) => {
         const id = consumeId++;
+        consumeId = consumeId % Number.MAX_SAFE_INTEGER;
         const task: [Resolve<A>, Reject, number] = [resolve, reject, id];
 
         consumer.push(task);
--- a/src/Lazy/concurrentPool.ts
+++ b/src/Lazy/concurrentPool.ts
@@ -68,7 +68,7 @@ function concurrentPool<A>(
       idleWorkers++;
 
       itemMap.set(id, item);
-      await chainItem(item);
+      chainItem(item);
     } catch (error) {
       idleWorkers++;
       finished = true;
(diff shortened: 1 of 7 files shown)


src/Lazy/concurrentPool.ts
```diff
@@ -0,0 +1,179 @@
...
+
+      itemMap.set(id, item);
+      await chainItem(item);
+    }
+  }
+
+  async function chainItem(item: Item<A>) {
+    prev = prev.then(() => item).then(pullItem);
+    return item;
+  }
```
hg-pyun, line 89: Is there any reason to use the `async` keyword if await is not used inside chainItem?

src/Lazy/concurrentPool.ts
```diff
@@ -0,0 +1,179 @@
...
+  success:
+    | { value: undefined; done: true } // done
+    | { value: T; done: false };
+};
+
+/**
+ * @experimental concurrentPool
+ */
+function concurrentPool<A>(
+  length: number,
```
hg-pyun, line 15: I’m not sure if the name length clearly represents its role. Is this a common variable name in functions dealing with concurrency?

hg-pyun (commented):
I think the concurrentPool is same here [code logic](https://github.com/marpple/DevUniv/blob/main/apps/live/src/pages/live-1/concurrent2.ts) that written by @indongyoo. Referencing this code might be helpful.

src/Lazy/concurrentPool.ts
ppeeou: I tried to match it with the variable name of [concurrent](https://github.com/marpple/FxTS/blob/main/src/Lazy/concurrent.ts#L66).

When changing it, it would be better to proceed with the variable name of concurrent!
ex) `concurrency`?

hg-pyun: got it

ysm-dev (approved):
LGTM

ysm-dev (approved):

src/Lazy/concurrentPool.ts
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * @example
+ * ```ts
+ * await pipe(
+ *   [1, 2, 3, 4, 5, 6],
+ *   toAsync,
+ *   map((a) => delay(100 * a, a)),
+ *   concurrentPool(3),
+ *   each(console.log), // log 1, 2, 3, 4, 5, 6
+ * ); // 2 seconds
```
ysm-dev, line 24: missing backticks 
https://fxts.dev/docs/concurrentPool
@ppeeou

Issue #305, opened 2025-02-27
Have you considered a function called `When`?


freevuehub:
# Suggestion

Have you considered a function called `When`?

It’s a function that returns either the second argument (a function or a value) if a given condition is met.
At first glance, it might seem similar to `throwIf`, but instead of throwing an error, it simply returns a value.

Additionally, if the condition is not met, it returns the received value as is, allowing it to be used in the next pipeline.

If this functionality already exists and I just missed it in the documentation, I apologize.

## ⭐ Suggestion

I’m currently using the following code as a temporary implementation.

```Typescript
import { isUndefined } from '@fxts/core'

function when<T>(predicate: (item: T) => boolean, callback: () => void): (iterator: T) => T
function when<T>(predicate: (item: T) => boolean, callback: () => void, iterator: T): T

function when<T>(predicate: (item: T) => boolean, callback: () => void, iterator?: T) {
  if (isUndefined(iterator))
    return (currentIterator: T) => when(predicate, callback, currentIterator)
  if (predicate(iterator)) return callback()

  return iterator
}

export default when
```

## 💻 Use Cases

```Typescript
pipe(
  event.clientX / window.innerWidth,
  when(
    (percent) => percent <= 0.2,
    () => // If the condition is met, execute and stop the pipeline.
  ),
  when(
    (percent) => percent >= 0.8,
    () => // If the condition is met, execute and stop the pipeline.
  ),
  (value) => {
    console.log(value) // event.clientX / window.innerWidth
  }
)
```

ppeeou:
I think it will be useful✨ @freevuehub 

Could you please upload what you suggested as a PR?

freevuehub:
@ppeeou 

Are there any mandatory tests that must be conducted or additional features that should be added?
I have completed the application tests for the pipe function.

ppeeou:
I think, as you mentioned above, it should be enough to test only the predicate true and false cases.

PR #309, opened 2025-03-09
feat: add `when`

freevuehub:
Fixes https://github.com/marpple/FxTS/issues/305

The initial when function was designed to always terminate the pipeline based on the given condition. However, after implementing it, I realized that it was functionally identical to throwIf.

To enhance its differentiation, I added the following features:

1. The first argument accepts a Boolean value.
2. If the function in the second argument returns a value, that value is passed to the next pipe. If it returns nothing, an error is thrown to terminate the pipeline.

If the existing throwIf function is deemed sufficient, the when function may not be necessary.


ppeeou:
@freevuehub  thank you for your contribution🎉

As you wrote in the issue, `when` seems to be useful just because it doesn't return an error if the predicate condition is met, unlike `throwIf`.
I think it would be good if you added it as you wrote in the issue!


I also request that you write a type test.
https://github.com/marpple/FxTS/tree/main/type-check

freevuehub:
@ppeeou 
I have written it.

src/when.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @example
+ * ```ts
+ * pipe(
+ *   100,
+ *   when(
+ *     isNumber,
+ *     () => `This is number`
+ *   )
+ * ) // This is number
```
ppeeou, line 14: Could you please add some generic examples, not just `pipe`?

freevuehub: @ppeeou 

I have added additional explanations to the document.

Additionally, I have included a test code matching the given example and added it to the existing pipe to verify that it produces the same result.

ppeeou (approved):
lgtm👍

PR #312, opened 2025-03-27
feat: change `takeUntil` to `takeUntilInclusive`

### LINKED ISSUE #311: feat: change the name of takeUntil to takeUntilInclusive.
# Suggestion

## ⭐ Suggestion

The existing `takeUntil` implementation in `fxts` returns the value when the predicate becomes true.
In contrast, the implementation of `takeUntil` in `rxjs` does not return the value corresponding to the true condition of the predicate. To avoid confusion with the usage of the existing `takeUntil`, we will rename `fxts`'s `takeUntil` to `takeUntilInclusive`.

https://rxjs.dev/api/index/function/takeUntil

ppeeou:
Fixes #311

c.c @indongyoo

### DIFF
--- a/src/Lazy/fx.ts
+++ b/src/Lazy/fx.ts
@@ -26,7 +26,7 @@ import peek from "./peek";
 import reject from "./reject";
 import slice from "./slice";
 import take from "./take";
-import takeUntil from "./takeUntil";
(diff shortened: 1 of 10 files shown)


hg-pyun:
Merged!

PR #323, opened 2025-06-13
fix: entries return type

ppeeou:
A situation where the type does not work based on the JavaScript runtime behavior.

Modified to transfer types based on the JavaScript runtime behavior.

Both string and number are entered as key values ​​in the object.
```
var obj = { 2025 : 'hello' }
obj[2025] // hello
obj['2025'] // hello
```

The results of `Object.keys(obj)`, `Object.entries`, and `for k in obj` are all strings, so if the first key of `entries` is a number, the type is transferred as string.

The existing response type is maintained(Maintaining backward compatibility)

### DIFF
--- a/src/Lazy/entries.ts
+++ b/src/Lazy/entries.ts
@@ -1,7 +1,14 @@
-type Entries<
-  T extends Record<string, any>,
-  K extends keyof T,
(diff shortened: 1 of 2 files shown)


hg-pyun (approved):
LGTM

PR #333, opened 2025-07-03
docs: edit example concurrentPool

odoldotol:
expected 900 ms. not 2 sec.

Fixes #


ppeeou (approved):
@odoldotol oh.. you're right, thank you

ppeeou:
Looking at the example, it seems like the expression is a bit difficult to predict the behavior..🧐
I'll have to think about it

PR #339, opened 2025-08-24
fix: resolve type inference issue with groupBy for enum and union types (#233 issue)

### LINKED ISSUE #233: Type inference is not working when used with Enum.
# Bug Report

### 💻 Code
https://codesandbox.io/s/competent-star-g4nqnp?file=/src/App.tsx

```ts

import { entries, fromEntries, groupBy, map, pipe, reduce } from "@fxts/core";
import "./styles.css";

type DataSet = {
  status: Status;
  count: number;
};

enum Status {
  Todo = "TODO",
  InProgress = "IN_PROGRESS",
  Done = "DONE"
}

const data: Array<DataSet> = [
  {
    status: Status.Todo,
    count: 0
  },
  {
    status: Status.InProgress,
    count: 0
  }
];

export default function App() {
  pipe(
    data,
    map((item) => {
      return {
        progress: Status.Done,
        count: item.count
      };
    }),
    groupBy((item) => item.progress),
    entries,
    map(([progress, values]) => {
      return [
        progress,
        reduce((acc, cur) => acc + cur.count, 0, values) // <-- Type Error: values
      ] as const;
    }),
    fromEntries
  );

  return null;
}

```

### 🙁 Actual behavior

TypeScript Inference is not working.

### 🙂 Expected behavior

TypeScript Inference works.

---

### Version Information

@fxts/core: 0.14.0: fine.
@fxts/core: 0.15.0: bug occur.

elrion018:
This PR fixes issue #233 where `groupBy` function causes type errors when used with enum and union types in combination with `entries` and `reduce`.

## Problem (Issue #233)

When using the `groupBy` function with enum or union types, excessive type narrowing caused type errors when combined with functions like `entries` and `reduce`.

Example code from the issue:
```typescript
enum Status {
  Todo = "TODO",
  InProgress = "IN_PROGRESS",
  Done = "DONE"
}

const data = [
  { progress: Status.Todo, count: 10 },
  { progress: Status.Done, count: 20 },
];

// Type error!
const result = pipe(
  data,
  groupBy((item) => item.progress),
  entries,
  map(([progress, values]) => [
    progress,
    reduce((acc, cur) => acc + cur.count, 0, values) // ❌ Type error
  ])
);
```

## Root Cause

The previous `groupBy` implementation created different specific types for each enum value through type narrowing:

```typescript
// Previous type inference result
{
  "TODO": ({ progress: Status; count: number } & { progress: Status.Todo })[];
  "IN_PROGRESS": ({ progress: Status; count: number } & { progress: Status.InProgress })[];
  "DONE": ({ progress: Status; count: number } & { progress: Status.Done })[];
}
```

After passing through `entries`, `values` becomes a union of multiple different array types:
```typescript
type ValuesType =
  | ({ progress: Status.Todo; count: number })[]
  | ({ progress: Status.InProgress; count: number })[]
  | ({ progress: Status.Done; count: number })[];
```

`reduce` expects an `Iterable<T>` where `T` is a specific type, but a union of arrays (`A[] | B[] | C[]`) represents "one of several array types" - TypeScript cannot determine which specific array type it is, making it impossible to safely type-check.

## Solution

Simplified the type definition of `groupBy` to remove unnecessary type narrowing:

**Before:**
```typescript
[K in B]: (A & { [K2 in GetKeyOf<A, B>]: K })[]
```

**After:**
```typescript
[K in B]: A[]
```

Now it returns arrays of the same type for all keys:
```typescript
// Improved type inference result
{
  "TODO": { progress: Status; count: number }[];
  "IN_PROGRESS": { progress: Status; count: number }[];
  "DONE": { progress: Status; count: number }[];
}

// Values type after entries
type ValuesType = { progress: Status; count: number }[];
// Single array type, so reduce works correctly
```

## When Type Narrowing is Needed

If users really need type narrowing, we recommend using explicit type guards or directly accessing each group:

```typescript
const grouped = groupBy((item) => item.status, data);

// Option 1: Use type assertion when directly accessing
const todoItems = grouped[Status.Todo] as Array<{ status: Status.Todo; count: number }>;

// Option 2: Use type guard functions
function isTodoItem(item: any): item is { status: Status.Todo; count: number } {
  return item.status === Status.Todo;
}

// Option 3: Process each group separately
if (Status.Todo in grouped) {
  const todos = grouped[Status.Todo];
  // Apply additional type narrowing if needed
}
```

## Supported Types

With this fix, `groupBy` works correctly with all of the following types:

- ✅ String enum: `enum Status { Todo = "TODO", ... }`
- ✅ Numeric enum: `enum Level { Low = 0, High = 1 }`
- ✅ String union: `type Color = "red" | "green" | "blue"`
- ✅ Number union: `type Level = 1 | 2 | 3`
- ✅ String literal: `"A" as const`
- ✅ Regular string: `string`

**Note on number union types:**
Number union types (`1 | 2 | 3`) work correctly with `groupBy` itself. However, the `entries` function only accepts `Record<string, any>` types, so objects with number keys cannot be used with `entries`. This is a separate issue from `groupBy`.

### Type tests:
- Added type inference tests for all data types including enum, string union, number union, and string literal
- Includes tests for the core `entries + reduce` pattern from issue #233

### Runtime tests:
- Added 6 new test cases
- Verified actual behavior with enum and union types
- All existing tests pass

## Impact

This change only affects type-level behavior and does not impact runtime behavior. It provides a better developer experience for most real-world usage patterns (especially the `entries + reduce` combination).

**Fixes #233**

### DIFF
--- a/test/groupBy.spec.ts
+++ b/test/groupBy.spec.ts
@@ -1,4 +1,15 @@
-import { filter, fx, groupBy, pipe, toAsync } from "../src";
+import {
+  entries,
+  filter,
+  fromEntries,
+  fx,
+  groupBy,
+  map,
+  pipe,
+  reduce,
+  toArray,
+  toAsync,
+} from "../src";
 import { AsyncFunctionException } from "../src/_internal/error";
 
 type Obj = {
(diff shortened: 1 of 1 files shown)


ppeeou:
@elrion018  Thank you for your interest in the library.

Before submitting the PR, I have a question.
- Is there any other reference code that provides the simple types for groupBy as shown in the PR?
- In the case of groupBy, is it more useful to support a wider type rather than using a narrow type?

The intention was to narrow the types for groupBy
Resolving the issue seems straightforward, but I think we need a rationale for our choice.

I'd also like to hear  your  opinion. @hg-pyun

elrion018:
@ppeeou 

I looked at how other popular functional libraries handle `groupBy` types:

**es-toolkit:**
```typescript
function groupBy<T, K extends PropertyKey>(
  arr: T[], 
  getKeyFromItem: (item: T) => K
): Record<K, T[]>
```

**fp-ts NonEmptyArray:**
```typescript
groupBy: <A>(f: (a: A) => string) => (as: A[]) => Record<string, NonEmptyArray<A>>
```

Both just return simple array types (`T[]` or `NonEmptyArray<A>`) - no complex intersection types or fancy type narrowing.

The `entries + reduce` pattern is super common when working with grouped data. If users can't do this basic workflow without hitting type errors, it really hurts the library's usability.

This change brings FxTS in line with what other libraries are doing, which feels like the right move.

PR #343, opened 2025-09-08
feat: add `resolveProps` (#340)

### LINKED ISSUE #340: feat: Resolve all props
# Suggestion

A function to resolve all of the props.

## ⭐ Suggestion

There are three implements.

- Using `Promise.all`: The most general implement.
  ``` ts
  const resolveProps: <T extends object>(obj: T) =>
  Promise<{ [K in keyof T]: Awaited<T[K]> }>
    = <T extends object>(obj: T) => 
      Promise.all(
        Object.entries(obj)
        .map(async ([k, v]) => [k, await v]),
      ).then(Object.fromEntries);
  ```
- Using `Array.fromAsync`: It looks like previous implement, but use`Array.fromAsync`. So it's lazy.
  ``` ts
  const resolveProps: <T extends object>(obj: T) =>
  Promise<{ [K in keyof T]: Awaited<T[K]> }>
    = <T extends object>(obj: T) => 
      Array.fromAsync(
        Object.entries(obj),
        async ([k, v]) => [k, await v],
      ).then(Object.fromEntries);
  ```
- Using functions in this library: It looks good, but has a type error.
  ``` ts
  const resolveProps: <T extends object>(obj: T) =>
  Promise<{ [K in keyof T]: Awaited<T[K]> }>
    = <T extends object>(obj: T) =>
      pipe(
        obj,
        entries,
        map(async ([k, v]) => [k, await v] as const),
        toAsync,
        fromEntries,
      );
  ```

  
## 💻 Use Cases

`resolveProps` handles object that has promise attributes.

And if this function is defined, then also define `asyncEvolve` simply
```ts
const asyncEvolve = (transformation) => (obj) =>
  pipe(obj, evolve(transformation), resolveProps)

const a = await asyncEvolve({ a: async (a) => a + 1 })({ a: 1 });
console.log(a); // { a: 2 }
```

2chanhaeng:
Fixes #340

### DIFF
--- a/src/resolveProps.ts
+++ b/src/resolveProps.ts
@@ -1,0 +1,53 @@
+/**
+ * Resolve all properties of an object that may be promises.
+ *
+ * @example
+ * ```ts
+ * const obj = {
+ *  a: Promise.resolve(1),
+ *  b: 2,
+ *  c: Promise.resolve(3),
+ * }
+ *
+ * const result = await resolveProps(obj);
+ * console.log(result); // { a: 1, b: 2, c: 3 }
+ * ```
+ *
+ * Also, with this function, you can define `asyncEvolve` simply.
+ *
+ * ```ts
+ * import { evolve, pipe, resolveProps } from "@fx-ts/core";
+ *
+ * const asyncEvolve = <T extends object>(transformation:Transformation<T>) =>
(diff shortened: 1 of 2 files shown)


ppeeou (approved):
Thank you for your contribution 🎉

PR #345, opened 2025-09-09
Fix `Entries` type (#342)

### LINKED ISSUE #342: Type error or wrong implementattion of `entries` with objects that has symbol keys.
# Bug Report

Type error or wrong implementattion of `entries` with objects that has symbol keys.

### 💻 Code

```ts
const a = entries({ "key": "value", [Symbol("key")]: "value" });
console.log([...a]);
```

### 🙁 Actual behavior

Image
Image
Image

`a` is inferred by tsc as `Generator<[string | symbol, string], void>`. However, when the actually output the values, only keys of type `string` remain.

### 🙂 Expected behavior

The generator type should include `string` only or the result generator include `symbol` too.

There are two options:
- Fixing `Entries` type to extract `string` keys. `Object.prototype.entries` also like this.
  ```diff
    type Entries<T extends Record<Key, any>, K extends keyof T> = K extends number
      ? [NumberToString<K>, T[K]]
  -   : K extends Key
  +   : K extends Extract<keyof T, string>
      ? [K, T[K]]
      : never;
  ```
- Fixing the `entries` function to yield `symbol` keys. I think this way is more useful. 
  ```diff
    function* entries<T extends Record<Key, any>>(
      obj: T,
    ): Generator<Entries<T, keyof T>, void> {
      for (const k in obj) {
        if (Object.prototype.hasOwnProperty.call(obj, k)) {
          yield [k, obj[k]] as any;
        }
      }
  +   yield* Object.getOwnPropertySymbols(obj).map((k) => [k, obj[k]] as any);
    }
  ```

If the second option is chosen, then it would be better to fix `keys` and `values`. They don't have type error now because they didn't have return type interface.

---

### Version Information

- Operating system: macOS 15.6
- Typescript: 5^
- Nodejs: 22.12.0 (`Symbol` and `Object.getOwnPropertySymbols` in Node.js from 0.12)

2chanhaeng:
Fixes #342

### DIFF
--- a/src/Lazy/entries.ts
+++ b/src/Lazy/entries.ts
@@ -6,7 +6,7 @@ type NumberToString<T extends number> = `${T}` extends infer R extends string
 
 type Entries<T extends Record<Key, any>, K extends keyof T> = K extends number
   ? [NumberToString<K>, T[K]]
-  : K extends Key
+  : K extends Extract<Key, string>
   ? [K, T[K]]
   : never;
 
--- a/type-check/entries.text.ts
+++ b/type-check/entries.text.ts
@@ -1,0 +1,15 @@
+import { entries } from "../src";
+import * as Test from "../src/types/Test";
+
+const { checks, check } = Test;
+
+const symbolKey = Symbol("a");
+const res1 = entries({ 1: 1, a: "a", [symbolKey]: "a" });
+
+checks([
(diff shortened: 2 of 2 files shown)


ppeeou (commented):

I left some comments

Issue #341, opened 2025-09-02
Do you have any plans to provide utility types as well?


freevue:
# Suggestion

## ⭐ Suggestion

Recently, I ran into an interesting case while developing.
I needed to convert an interface written in camelCase format into UPPER_SNAKE_CASE format.

Until now, I’ve been declaring new types or creating small utility helpers myself to handle these cases.
That got me wondering—are there any libraries that already provide such utility types? And at the same time, I thought maybe fxts could also provide them.

I was curious if you have any plans to include this kind of functionality.
I’d also like to suggest it, since I think supporting these kinds of utility types could be fun and useful.

## 💻 Use Cases

```typescript
type UpperSnake<S extends string> = S extends `${infer A}${infer B}`
  ? `${A extends Uppercase<A> ? '_' : ''}${Uppercase<A>}${UpperSnake<B>}`
  : Uppercase<S>

type Camel = 'helloWorld'
type Snake = UpperSnake<Camel> // HELLO_WORLD
```

ppeeou:
@freevue 
Sounds like a good idea.

It would be nice if it were provided in a `Type`  folder!

<img width="556" height="518" alt="Image" src="https://github.com/user-attachments/assets/f607f3af-4cc3-4e63-8a3a-61fcbafa0252" />

PR #348, opened 2025-09-14
feat: type uppercase

freevue:
Fixes [#341](https://github.com/marpple/FxTS/issues/341)

---

Currently, I’ve only implemented the process of converting from camelCase to UPPERCASE.
I also added a separator option so that the words can be split in between.

Since this part only involves types, I’ve only written type-checking tests.
For now, a new directory has been added, so I haven’t exported it to the main root yet.

At the moment, I’ve only added the functionality I personally needed, but I’ll continue improving it or adding more if there are other suggestions or directions.


hg-pyun:
t’s a good idea, but I’m not sure whether this feature really belongs in a functional programming library. If I were choosing a library for my project, I would probably have used [type-fest](https://github.com/sindresorhus/type-fest?tab=readme-ov-file) in this case. We should think carefully about whether this functionality is truly necessary in the context of the library we provide.

freevue:
@ppeeou @hg-pyun 

In that case, I’ll cancel this PR and close the issue.
I’ll contribute again in the future with different content.

Also, I’ll make sure to approach things a bit more carefully going forward.

PR #351, opened 2025-09-18
feat: add `match` and `cases` (#346)

### LINKED ISSUE #346: function: `match`
# Suggestion

## ⭐ Suggestion

It's a some kind of `when` extension.

```ts
match(
  cond1, then1,
  cond2, then2,
  ....,
  // default=identify
)
```

```ts
const match = (...fns: Function[]) => function <T>(x: T){
  for(const [cond, fn] of chunk(2, fns)) {
    if(!fn) return cond(x); // if `fns` has an odd number of item, the last function is default
    if(cond(x)) return fn(x);
  }
  return x // if `fns` has an even number of item, just return arg 
}
```

The `Function[]` is not good type, so the real implement should overwrite interface by every case like `pipe`.

## 💻 Use Cases

```ts
const users = [
  { name: "Kim", gender: "male" },
  { name: "Lee", gender: "female", grade: "doctor" },
  { name: "Park", gender: null },
];

console.log(
  pipe(
    users,
    map(match(
      (user) => user.grade === "doctor", ({ name }) => `Dr. ${name}`,
      (user) => user.gender === "female", ({ name }) => `Ms. ${name}`,
      (user) => user.gender === "male", ({ name }) => `Mr. ${name}`,
      ({ name }) => `Dear ${name}`,
    )),
    toArray,
  ),
); // ["Mr. Kim", "Dr. Lee", "Dear Park"]

const data = [1, "a", { a: "1" }];
console.log(
  pipe(
    data,
    match(
      isNumber, String,
      isObject, JSON.stringify,
    )
    toArray,
  )
); // ["1", "a", "{ a: "1" }"]
```

2chanhaeng:
Fixes #346

I also implement [`cases`](https://github.com/marpple/FxTS/issues/346#issuecomment-3289138628), more flexible than match.
If you don't think `cases` is needed, I'll reqeust again without `cases`.

### DIFF
--- a/src/cases.ts
+++ b/src/cases.ts
@@ -1,0 +1,231 @@
+import { chunk } from "./Lazy";
+
+type Predicate<T> = (value: T) => boolean;
+type Refinement<T, T1 extends T> = (value: T) => value is T1;
+type Mapper<T, R> = (value: T) => R;
+type MapperByRefinement<Pred, R> = Pred extends (value: infer T) => boolean
+  ? Pred extends Refinement<T, infer T1>
+    ? Mapper<T1, R>
+    : Mapper<T, R>
+  : never;
+type Thrower<T> = (error: T) => never;
+type DefaultFn<T, R> = Mapper<T, R> | Thrower<T>;
+type EvenArray = unknown[] & { length: 0 | 2 | 4 | 6 | 8 | 10 | 12 };
+
+function cases(): (value: unknown) => typeof value;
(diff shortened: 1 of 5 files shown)


2chanhaeng:
@ppeeou Fixed!

ppeeou (commented):
@2chanhaeng  Sorry for the late reply🙏

I think I'm missing the general test case writing part. Just adding that would be enough.

https://github.com/marpple/FxTS/tree/main/test

PR #352, opened 2025-09-21
fix: resolve type inference issue with indexBy for enum and union types

elrion018:
## Summary

This PR fixes the same type narrowing issue that was resolved for `groupBy` in PR #339, but for the `indexBy` function. The excessive type narrowing in `indexBy` caused type errors when using the results with `Object.entries()`, `Object.keys()`, `Object.values()`, and other common Object methods for enum and union types.

## Problem

The previous `indexBy` implementation used excessive type narrowing:

```typescript
[K in B]: A & { [K2 in GetKeyOf<A, B>]: K };
```

This created different specific types for each enum value, making it difficult to use the result with standard Object methods:

```typescript
enum UserRole {
  Admin = "admin",
  User = "user",
  Guest = "guest"
}

const users = [
  { id: 1, name: "Alice", role: UserRole.Admin },
  { id: 2, name: "Bob", role: UserRole.User }
];

const usersByRole = indexBy(user => user.role, users);

// ❌ Type error before this fix
Object.entries(usersByRole).forEach(([role, user]) => {
  console.log(`${role}: ${user.name}`); // user type inference failed
});
```

## Solution

Simplified the type definition to remove unnecessary type narrowing:

```typescript
[K in B]: A;
```

Now it returns values of the same type for all keys, making Object methods work correctly:

```typescript
// ✅ Works correctly after this fix
Object.entries(usersByRole).forEach(([role, user]) => {
  console.log(`${role}: ${user.name}`); // user type correctly inferred
});
```

## Supported Types

With this fix, `indexBy` now works correctly with:

- ✅ String enum: `enum Status { Todo = "TODO", ... }`
- ✅ Numeric enum: `enum Level { Low = 0, High = 1 }`
- ✅ String union: `type Color = "red" | "green" | "blue"`
- ✅ String literal: `"A" as const`
- ✅ Regular string: `string`

### Runtime tests:
- Added 6 new test cases covering enum, union, and literal types
- Verified `Object.entries()`, `Object.keys()`, `Object.values()` patterns work
- All existing tests continue to pass

### Type tests:
- Added type inference tests for enum and union types
- Verified Object method patterns type-check correctly

## Impact

This change only affects type-level behavior and does not impact runtime behavior. All existing code continues to work, but previously failing type patterns now work correctly.

Fixes the same class of issues as PR #339 but for the `indexBy` function.

### DIFF
--- a/src/indexBy.ts
+++ b/src/indexBy.ts
@@ -46,14 +46,14 @@ function indexBy<A extends object, B extends Key & A[keyof A]>(
   f: (a: A) => B,
   iterable: Iterable<A>,
 ): {
-  [K in B]: A & { [K2 in GetKeyOf<A, B>]: K };
+  [K in B]: A;
 };
 
 function indexBy<A extends object, B extends Key & A[keyof A]>(
   f: (a: A) => B | Promise<B>,
   iterable: AsyncIterable<A>,
 ): Promise<{
-  [K in B]: A & { [K2 in GetKeyOf<A, B>]: K };
+  [K in B]: A;
 }>;
 
 function indexBy<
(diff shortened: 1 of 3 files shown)


ppeeou:
👍

PR #362, opened 2025-12-24
fix: support map and set data structure

### LINKED ISSUE #361: Object iteration functions do not work with Map and Set objects
# Bug Report

### 💻 Code

```js
import { entries, keys, values, pick, omit } from "@fxts/core";

const map = new Map([["a", 1], ["b", 2], ["c", 3]]);

console.log([...entries(map)]);  // []
console.log([...keys(map)]);     // []
console.log([...values(map)]);   // []
console.log(pick(["a"], map));   // {}

// Expected behavior with native methods
console.log([...map.entries()]); // [["a", 1], ["b", 2], ["c", 3]]
console.log([...map.keys()]);    // ["a", "b", "c"]
console.log([...map.values()]);  // [1, 2, 3]
```

> **Note:** TypeScript type signatures currently don't accept Map/Set, but the runtime behavior is also broken when types are bypassed.

### 🙁 Actual behavior

Nine functions return empty results when given `Map` or `Set` objects instead of iterating their entries.

<details open>
<summary><strong>Affected Functions</strong></summary>

**Iterator functions (using `for...in` loop):**
- `entries` - `src/Lazy/entries.ts:31-35`
- `keys` - `src/Lazy/keys.ts:13-17`
- `values` - `src/Lazy/values.ts:13-17`

**Transformation functions (using `Object.entries()`):**
- `pick` - `src/pick.ts:7`
- `omit` - `src/omit.ts:7`
- `pickBy` - `src/pickBy.ts:80`
- `omitBy` - `src/omitBy.ts:80`
- `evolve` - `src/evolve.ts:50`
- `compactObject` - `src/compactObject.ts:65`

</details>

<details>
<summary><strong>Root Cause Analysis</strong></summary>

Both `for...in` loops and `Object.entries()` only iterate over enumerable string properties:

```ts
// Current implementation pattern
function* entries(obj) {
  for (const k in obj) {  // Only iterates enumerable properties
    if (Object.prototype.hasOwnProperty.call(obj, k)) {
      yield [k, obj[k]];
    }
  }
}
```

Map/Set objects don't expose their data as enumerable properties, so these loops find nothing to iterate.

</details>

### 🙂 Expected behavior

These functions should work with Map/Set objects by utilizing their native iteration methods (`.entries()`, `.keys()`, `.values()`).

**Note:** `isEmpty` already handles Map/Set correctly via `instanceof` checks at `src/isEmpty.ts:38`.

### Proposed Solution

<details open>
<summary><strong>Group 1: Iterator Functions (entries, keys, values)</strong></summary>

Add type guards to detect Map/Set and delegate to native methods:

```diff
 function* entries<T extends Record<Key, any>>(
   obj: T,
 ): Generator<Entries<T, keyof T>, void> {
+  if (obj instanceof Map || obj instanceof Set) {
+    yield* obj.entries();
+    return;
+  }
   for (const k in obj) {
     if (Object.prototype.hasOwnProperty.call(obj, k)) {
       yield [k, obj[k]] as any;
     }
   }
 }
```

Same pattern for `keys()` and `values()`:
```ts
if (obj instanceof Map || obj instanceof Set) {
  yield* obj.keys();    // for keys()
  yield* obj.values();  // for values()
  return;
}
```

</details>

<details open>
<summary><strong>Group 2: Transformation Functions (pick, omit, pickBy, omitBy, evolve, compactObject)</strong></summary>

Convert Map entries to array, process, and return as object:

```diff
 function pick<T extends object, K extends keyof T>(
   keys: readonly K[],
   obj: T,
 ): Pick<T, K> {
+  if (obj instanceof Map) {
+    const entries = [...obj].filter(([k]) => keys.includes(k));
+    return Object.fromEntries(entries);
+  }
   return Object.fromEntries(
     Object.entries(obj).filter(([k]) => keys.includes(k as K))
   ) as Pick<T, K>;
 }
```

This approach:
- Maintains backward compatibility (returns plain objects)
- Works consistently with existing type signatures
- Leverages native Map iteration

</details>

---

### Version Information

- Operating system: macOS
- Typescript: 5.x
- Node.js: 22.x

---

### Appendix

[Issue Reporting Session](https://opncd.ai/share/yZO5IGz0)

hg-pyun:
Fixed a bug where the entries function did not support Map and Set. While the same issue exists for keys and values, it is not addressed in this PR. The issue was caused by using a for...in loop, which does not work with Map and Set. When an iterable object is used in a for...in loop, it returns `undefined`.

## Reference
- Fixes #361 
- https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Set/entries

### DIFF
--- a/src/Lazy/entries.ts
+++ b/src/Lazy/entries.ts
@@ -4,33 +4,65 @@ type NumberToString<T extends number> = `${T}` extends infer R extends string
   ? R
   : never;
 
-type Entries<T extends Record<Key, any>, K extends keyof T> = K extends number
+type ObjectEntries<
+  T extends Record<Key, any>,
+  K extends keyof T,
+> = K extends number
   ? [NumberToString<K>, T[K]]
   : K extends string
   ? [K, T[K]]
   : never;
 
+type EntriesResult<T> = T extends Map<infer K, infer V>
+  ? Generator<[K, V], void>
+  : T extends Set<infer V>
+  ? Generator<[V, V], void>
+  : T extends Record<Key, any>
+  ? Generator<ObjectEntries<T, keyof T>, void>
+  : never;
+
 /**
  *
- * Returns an iterator of the own enumerable string keyed-value pairs.
+ * Returns an iterator of key-value pairs.
+ *
+ * Supports plain objects, Map, and Set:
+ * - For objects: returns own enumerable string keyed-value pairs
+ * - For Map: returns [key, value] pairs
+ * - For Set: returns [value, value] pairs
  *
  * @example
  * ```ts
- *
  * [...entries({ a: 1, b: "2", c: true })]
  * // [["a", 1], ["b", "2"], ["c", true]]
- * ```
  *
+ * [...entries(new Map([["a", 1], ["b", 2]]))]
+ * // [["a", 1], ["b", 2]]
+ *
+ * [...entries(new Set([1, 2, 3]))]
+ * // [[1, 1], [2, 2], [3, 3]]
+ * ```
  *
  * see {@link https://fxts.dev/docs/fromEntries | fromEntries}
  */
 
-function* entries<T extends Record<Key, any>>(
+function entries<T extends Map<any, any> | Set<any> | Record<Key, any>>(
   obj: T,
-): Generator<Entries<T, keyof T>, void> {
-  for (const k in obj) {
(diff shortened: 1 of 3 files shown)


ppeeou (approved):
LGTM 👍👍

PR #367, opened 2025-12-30
fix: restore test cases accidentally deleted

hg-pyun:
#365 

Restore test cases accidentally deleted in the issue above.


ppeeou (approved):
lgtm

PR #385, opened 2026-06-17
fix: remove any type in concurrent with functions

hg-pyun:
### Summary
  The concurrency marker that `concurrent(n)` threads back up the lazy chain travels through the iterator protocol's `next(value)` argument (the
  generator `TNext` slot). That channel was completely untyped — `Concurrent.of(length) as any` at the source and `_concurrent: any` in ~20
  operators. This PR types it properly with no runtime change.

  ### What changed
  - Added `export type ConcurrentArg = Concurrent | undefined` in `concurrent.ts` to document/name the channel.
  - Receiving side: `next(_concurrent: any)` → `next(_concurrent?: Concurrent)`.
  - Forwarding side: upstream iterators annotated as `AsyncIterator<A, unknown, ConcurrentArg>` so the marker forwards without a cast.
  - Removed the source cast (`Concurrent.of(length) as any` → `Concurrent.of(length)`) and the two casts in `filter` (`concurrent as any`,
  `recur(concurrent as Concurrent)`) by widening internal helpers to `ConcurrentArg`.
  - Same fix applied to `fork`'s internal `next` channel.

  Public return types (`IterableIterator` / `AsyncIterableIterator`) are unchanged — assignability is preserved via method bivariance, so this
  is type-only with zero behavior change.

  ### Why it's better than `any`
  `any` disabled checking on the marker and everything it touched. Now the compiler:
  - **enforces the `isConcurrent` guard** — `_concurrent.length` without narrowing is a compile error (`possibly 'undefined'`), preventing
  `undefined` deref;
  - **verifies the forwarding chain** — passing the marker to an iterator that can't receive it is now caught, instead of silently degrading
  concurrency to sequential;
  - **rejects wrong values** (e.g. `next(42)`) and **catches typos** on the marker;
  - restores autocomplete / hover / safe rename for the channel.

  ```ts
  // before: next(_concurrent: any)
  _concurrent.length            // ✅ compiles, may crash at runtime
  upstream.next(_concurrent)    // ✅ compiles, even if upstream can't receive it

  // after: next(_concurrent?: Concurrent)
  _concurrent.length            // ❌ TS18048: '_concurrent' is possibly 'undefined'
  if (isConcurrent(_concurrent)) _concurrent.length  // ✅ narrowed

### DIFF
--- a/src/Lazy/append.ts
+++ b/src/Lazy/append.ts
@@ -1,7 +1,11 @@
 import { isAsyncIterable, isIterable, isPromise } from "../_internal/utils";
 import type Awaited from "../types/Awaited";
 import type ReturnIterableIteratorType from "../types/ReturnIterableIteratorType";
-import concurrent, { isConcurrent } from "./concurrent";
+import concurrent, {
+  isConcurrent,
+  type Concurrent,
+  type ConcurrentArg,
+} from "./concurrent";
 
 function* sync<A>(a: A, iterable: Iterable<A>) {
   yield* iterable;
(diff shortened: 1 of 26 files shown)


ppeeou (approved):
lgtm 🚀🚀

